Issue #8, opened 2017-01-13
undefined reference to `app_main'


mintisan:
OS : macOS 10.11.6
PlatformIO: 1.7.1(IDE), 3.2.1(Core)

use [espidf-hello-world](https://github.com/platformio/platform-espressif32/tree/develop/examples/espidf-hello-world) example, then come error below(I can use Arduino example normally): 

```
Indexing .pioenvs/esp32dev/libtcpip_adapter.a
Linking .pioenvs/esp32dev/firmware.elf
.pioenvs/esp32dev/libesp32.a(cpu_start.o):(.literal.main_task+0x0): undefined reference to `app_main'
.pioenvs/esp32dev/libesp32.a(cpu_start.o): In function `main_task':
/Users/linjinhui/.platformio/packages/framework-espidf/components/esp32/cpu_start.c:169: undefined reference to `app_main'
collect2: error: ld returned 1 exit status
*** [.pioenvs/esp32dev/firmware.elf] Error 1

========================== [ERROR] Took 15.54 seconds ==========================
```

mintisan:
should add `extern "C"` in front of `app_main`

```
/* Hello World Example
   This example code is in the Public Domain (or CC0 licensed, at your option.)
   Unless required by applicable law or agreed to in writing, this
   software is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
   CONDITIONS OF ANY KIND, either express or implied.
 */
#include <stdio.h>
#include "freertos/FreeRTOS.h"
#include "freertos/task.h"
#include "esp_system.h"
#include "nvs_flash.h"

void hello_task(void *pvParameter)
{
        printf("Hello world!\n");
        for (int i = 10; i >= 0; i--) {
                printf("Restarting in %d seconds...\n", i);
                vTaskDelay(1000 / portTICK_RATE_MS);
        }
        printf("Restarting now.\n");
        fflush(stdout);
        system_restart();
}

extern "C" void app_main()
{
        nvs_flash_init();
        system_init();
        xTaskCreate(&hello_task, "hello_task", 2048, NULL, 5, NULL);
}

```

Gryffoon:
I had the same issue even if I've added an extern "C" 
with no result.

The problem was that the component.mk  was at the wrong place. It must be under the main folder. And not near to 'Makefile' file.

Buckler89:
@Navis-Raven How do you use the comoponent.mk in platformio? For me seems that platformio just ignore it.

[...]

PR #9, opened 2017-01-13
fix can't find app_main issue

mintisan:
can't compile successfully in macOS without `extern "C" `

### DIFF
--- a/examples/espidf-hello-world/src/main.c
+++ b/examples/espidf-hello-world/src/main.c
@@ -24,7 +24,7 @@ void hello_task(void *pvParameter)
     system_restart();
 }
 
-void app_main()
+extern "C" void app_main()
 {
     nvs_flash_init();
     system_init();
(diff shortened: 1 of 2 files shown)


mintisan:
for issue https://github.com/platformio/platform-espressif32/issues/8

ivankravets:
See https://travis-ci.org/platformio/platformio-core/jobs/191615208#L113

mintisan:
Is there some different on environment?

It works fine on my machine to add `extern "C"` before `void app_main()`.

```
src/main.c:27:8: error: expected identifier or '(' before string constant
extern "C" void app_main()
```

ivankravets:
Try to rename main.c to main.cpp

mintisan:
I commit a new https://github.com/platformio/platform-espressif32/pull/14 which fixed the error. And it work fine.

PR #14, opened 2017-01-17
fix can't find app_main issue with extern "C" and for c89 standard

mintisan:


ivankravets:
@valeros , could you review this PR?

ivankravets:
Obsolete PR with conflicted files.

PR #40, opened 2017-09-01
add basic support for custom partition table

drorgl:
add basic support for custom partition table, which should be in src/partitions_table.csv


ivankravets:
See http://docs.platformio.org/en/latest/platforms/espressif32.html#partition-tables

drorgl:
Much better, thank you!

PR #61, opened 2017-12-29
Add Lolin32-lite board

henricazottes:
Hi,

I'm using a variant of the Lolin32 board, called the "Lite" version (only the builtin led change as far as I tested it). Would be great to add it :)

Thanks !

### DIFF
--- a/boards/lolin32-lite.json
+++ b/boards/lolin32-lite.json
@@ -1,0 +1,33 @@
+{
+  "build": {
+    "core": "esp32",
+    "extra_flags": "-DARDUINO_LOLIN32_LITE",
+    "f_cpu": "240000000L",
+    "f_flash": "40000000L",
+    "flash_mode": "dio",
+    "ldscript": "esp32_out.ld",
+    "mcu": "esp32",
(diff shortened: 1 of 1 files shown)


ivankravets:
Hm... You need to make PR to https://github.com/espressif/arduino-esp32 with a custom "variant".

Otherwise, you can place a custom board manifest to http://docs.platformio.org/en/latest/platforms/creating_board.html#installation

henricazottes:
@ivankravets thanks for the feedback !

I've already made the PR on the esp32-arduino but it's not merged yet. As for the custom board, I wanted to share it as I may not be the only one using the Lolin32 lite board. Any reason why it wouldn't be merged here ?

ivankravets:
Because we depend on https://github.com/espressif/arduino-esp32
As result, PIO will fail on non existing variant

henricazottes:
Ok so the json board files are automatically generated from what is in the Expressif repo ?

It's just that I had to manually add it locally, I though I had to PR it here so it would be integrated in PIO.

PR #67, opened 2018-02-05
Set LED_BUILTIN if it is not defined by Arduino framework

sria91:
Set `LED_BUILTIN` if not already defined in `pins_arduino.h`.


ivankravets:
Thanks!

Issue #81, opened 2018-04-20
Some board files use '-' (dash) in macro names, but it is not a valid character for a macro name


prices:
There are a number of board files that set a macro with '-' in the macro name.  It causes this warning:

<command-line>:0:14: warning: ISO C++11 requires whitespace after the macro name

If you actually try to use the macro in a program, for instance:

#ifdef ARDUINO_ESP32-EVB
// Stuff here
#endif

It causes a further warning:

src/main.cpp:5:21: warning: extra tokens at end of #ifdef directive

The board files that are affected are:

esp32-evb.json:    "extra_flags": "-DARDUINO_ESP32-EVB",
esp32-gateway.json:    "extra_flags": "-DARDUINO_ESP32-GATEWAY",
esp32vn-iot-uno.json:    "extra_flags": "-DARDUINO_esp32vn-iot-uno",
m5stack-core-esp32.json:    "extra_flags": "-DARDUINO_M5Stack-Core-ESP32",
nodemcu-32s.json:    "extra_flags": "-DARDUINO_NodeMCU-32S",

I will create a pull request changing the '-' (dash) to '_' (underscore).  If that is not the solution you want, please let me know and I will change it.

ivankravets:
Resolved in https://github.com/platformio/platform-espressif32/pull/82

hongquan:
There is problem, which seems to be caused by this fix:

I'm using "esp32vn-iot-uno" board, now PlatformIO fails to build my code, with this error:

```
/home/quan/.platformio/packages/framework-arduinoespressif32/cores/esp32/Arduino.h:188:26: fatal error: pins_arduino.h: No such file or directory
```

Looking into the generated C compiler command:

```
xtensa-esp32-elf-gcc ... -DARDUINO_esp32vn_iot_uno ... -I/home/quan/.platformio/packages/framework-arduinoespressif32/variants/esp32vn_iot_uno
```

I notice that the included path is wrong:

```
/home/quan/.platformio/packages/framework-arduinoespressif32/variants/esp32vn_iot_uno
```
while it should be

```
/home/quan/.platformio/packages/framework-arduinoespressif32/variants/esp32vn-iot-uno
```

ivankravets:
@hongquan Thanks! Fixed. Please retest with http://docs.platformio.org/en/latest/platforms/espressif32.html#upstream

hongquan:
Thanks, @ivankravets . It works!

PR #82, opened 2018-04-20
The '-' (dash) has been changed to '_' (underscore) in the macro names.

prices:
This is a pull request for bug #81


ivankravets:
Thanks!

PR #113, opened 2018-07-16
Add ALKS ESP32 board variant

CornyjK:
As title sais.


CornyjK:
Could you check it @ivankravets ?

ivankravets:
Oh, sorry, I missed it :(

Thank you so much for the PR!

We will make the final release soon.

PR #142, opened 2018-11-23
Adding ability to include arduino-esp32 as component for espidf builds

### LINKED ISSUE #24: Arduino.h not found - ESP32 with ESP-IDF Framework
Hi,
Copying issue from https://community.platformio.org/t/arduino-h-not-found-esp32-with-esp-idf-framework/1257 to platform-espressif32.

Currently I am using an ESP32 with the IDF Framework.
According to the docs https://github.com/espressif/arduino-esp3217 (section: Using as ESP-IDF component) the IDF Framework (supported by Platform IO and set up in the platformio.ino file) has an Arduino Core implemented and can be called by the ´initArduino()´ function. But compiler throws an error that Arduino.h cannot be found.

Has this been solved? if not, hope somebody can help here.

------------- Log Start -----------------
Processing esp32dev (platform: espressif32; lib_ldf_mode: deep; board: esp32dev; framework: espidf)
 
Verbose mode can be enabled via `-v, --verbose` option
Collected 0 compatible libraries
Looking for dependencies...
Project does not have dependencies
Compiling .pioenvs\esp32dev\src\main.o

Generating partitions .pioenvs\esp32dev\partitions_table.bin
Generating LD script .pioenvs\esp32dev\esp32_out.ld
Compiling .pioenvs\esp32dev\bootloader\bootloader_start.o
Compiling .pioenvs\esp32dev\bootloaderSupport\src\bootloader_flash.o
Compiling .pioenvs\esp32dev\bootloaderSupport\src\bootloader_random.o
Compiling .pioenvs\esp32dev\bootloaderSupport\src\efuse.o
Compiling .pioenvs\esp32dev\bootloaderSupport\src\esp_image_format.o
Compiling .pioenvs\esp32dev\bootloaderSupport\src\flash_encrypt.o
Compiling .pioenvs\esp32dev\bootloaderSupport\src\flash_partitions.o
src\main.c:8:52: fatal error: components/arduino/cores/esp32/Arduino.h: No such file or directory
compilation terminated.
*** [.pioenvs\esp32dev\src\main.o] Error 1
------------- Log End ------------------

Regards,
Andy

atanisoft:
fixes #24.

By adding the following to the sdkconfig.h in the project using framework=espidf the framework-arduinoesp32 code is automatically included/compiled.
```
#define ENABLE_ARDUINO_DEPENDS 1
```

The default configuration is to require the application to define app_main as per usual with ESPIDF and call to arduinoInit(). By adding the following:
```
#define CONFIG_AUTOSTART_ARDUINO 1
```
the arduino-esp32 app_main will instead be used and the application can instead use setup()/loop() which are more common in arduino applications.


ionut-movila:
This saved my day, and hours of frustration trying to make countless changes just to include all the sources from Arduino in lib folder of the project. That ain't gonna work!

ionut-movila:
For other that don't know how to use it: until this very useful PR is merged you can just change in platformio.ini 

`
; platform = espressif32
platform = https://github.com/atanisoft/platform-espressif32#espidf-arduino
`

alvarolb:
With the latest version 1.6.0, how can I disable the CONFIG_AUTOSTART_ARDUINO in projects? 
I need to define my own app_main. Thanks!

atanisoft:
@alvarolb, you would do this in the menuconfig utility or by directly editing the sdkconfig.h that it generates. Also be sure to start with the latest sdkconfig.h provided in the arduino-esp32 code base under tools/sdk. There are a few config options that are required for some libraries to successfully compile (http client with ssl as example).

alvarolb:
Thanks @atanisoft, got it working by modifying the sdkconfig.h.

alvarolb:
But, just a question @atanisoft, it was working previously in 1.5.0 without modifying the sdkconfig.h. It also works out of the box in Arduino if I define the app_main. What changed? It is a problem to force users to modify the setup.. or to maintain it in future revisions.. Thanks!

atanisoft:
No clue, I haven't looked at the 1.6.0 code so there could be a bug introduced that will require updates to this PR. I'll take a look at it when I can.

atanisoft:
> it was working previously in 1.5.0 without modifying the sdkconfig.h. It also works out of the box in Arduino if I define the app_main. What changed? It is a problem to force users to modify the setup.. or to maintain it in future revisions.. Thanks!

I can't find anything that would have been broken from v1.5.0 to v1.6.0 (or v1.7.0 for that matter). Currently this depends on parsing the sdkconfig.h. It may make sense to shift this to parse the sdkconfig file instead but that would be a bit more of an involved change and I'm not sure of the value add in doing so. If you run "make menuconfig" it should be generating the sdkconfig.h automatically and inside menu config (with Arduino components included!) you can deselect the auto-start feature which will allow you to define your own app_main (the example uses this option as well).  As an alternative you could leverage the "void init()" or "void initVariant()" weakly defined functions from the arduino-esp32 project which get called as part of the arduinoInit() function (at the end), the default implementation for these are blank.

dralves:
This is extremely useful, thanks @atanisoft 
A couple of things I noticed:
- ENABLE_ARDUINO_DEPENDS is CONFIG_ENABLE_ARDUINO_DEPENDS as set by make menuconfig.
- the builder script is counting on finding ENABLE_ARDUINO_DEPENDS before CONFIG_AUTOSTART_ARDUINO. If it finds the latter first, then when it finds ENABLE_ARDUINO_DEPENDS it sets build_arduino_main to False, again

dralves:
The only "tricky" thing about this, is that it requires that we have installed "framework-arduinoespressif32" beforehand, offline. I came up with a variation where I actually added arduino as a component to esp-idf (my fork) and then made this script refer to that, if configured. Happy to contribute it back if people find value.

atanisoft:
The only way this plugin works is if you have the Espressif32 frameworks installed, without this the build py script isn't even available in PIO.

However, if you have a set of diffs to make this work better feel free to send them over and I'll update this PR with them.

VCO34:
Hi, I'm trying to run a simple example with I2C bus.I have a problem with the arduino libraries "Wire.h".

The example works when using the classic arduino framework, but I would like to use framework espidf with arduino lib.

Builder reported undefined references class TwoWire when linking. E.g. TwoWire :: read (), TwoWire :: beginTransmission (int), etc.

Platformio 2.0.0. Core 3.6.5

My settings plarformio.ini:

    [env: espidf-arduino]
    platform = https://github.com/atanisoft/platform-espressif32#espidf-arduino
    board = esp32dev
    framework = espidf
    monitor_speed = 115200

    build_flags =
      -D CONFIG_ESP_LYRAT_V4_3_BOARD
      -Ic: /Users/coufal/.platformio/packages/framework-arduinoespressif32@2.10000.0/libraries/Wire/src/

     lib_archive = false

main.cpp:
    
    #include "freertos/FreeRTOS.h"
    #include "freertos/task.h"
    #include "esp_system.h"
    #include "esp_log.h"
    #include "sdkconfig.h"
    #include <Arduino.h>
    #include "Adafruit_NFCShield_I2C.h"
    
    external "C" {
      void app_main ();
    }

    void app_main () {
      initArduino ();

     Adafruit_NFCShield_I2C nfc (IRQ, RESET);
     ...
    }

image


Thank you very much for your help.

Next,
Maybe I found a bug in compiling the program.
When builder build "eps_out.ld" in folder .pioenvs, the comment line is inserted as 

    /*
     *
     * Espressif IoT Development Framework Configuration
     *
     */
    // include arduino-esp32 libraries
    /* If BT is not built at all */
    MEMORY
    {

should be 

     /*include arduino-esp32 libraries*/

otherwise the compile will end up with an error.

Next question,
When it will be confirm pull request this feature to branche master ?

atanisoft:
@VCO34 the adafruit library you are using is not compatible with the esp32 libraries. You will need to find an alternative version of that library that supports the esp32. You will likely find more assistance with that library directly from the adafruit GitHub page rather than here.

Also you should not need to set a build_flags option for include path. You can specify it as ```lib_deps=Wire``` and PIO should set it automatically for you.

As for the ld file, that may be due to your inclusion of lib_archive in your PlatformIO.ini file. The builder expects to generally create a lib file for the arduino-esp32 code. I'm also not sure where that ld file is generated but I'll do some digging into it and correct that if I can.

atanisoft:
> * ENABLE_ARDUINO_DEPENDS is CONFIG_ENABLE_ARDUINO_DEPENDS as set by make menuconfig.

You are correct but the sdkconfig.h lists it as only ENABLE_ARDUINO_DEPENDS. In either case though it will match as the script is searching for a substring in the line and not an exact whitespace delimited word.

> * the builder script is counting on finding ENABLE_ARDUINO_DEPENDS before CONFIG_AUTOSTART_ARDUINO. If it finds the latter first, then when it finds ENABLE_ARDUINO_DEPENDS it sets build_arduino_main to False, again

Are you on the latest version of the code in this PR? In an earlier version it was resetting the variables incorrectly as you state but in the latest version that was corrected so that the order is not critical but instead just that they are present to enable the functionality.

kevinresol:
Getting this error:
```
~/.platformio/packages/framework-arduinoespressif32/cores/esp32/Esp.cpp:160:59: error: 'esp_image_verify' was not declared in this scope
```

To reproduce:
1. in `~/.platformio/packages`, delete `framework-arduinoespressif32` and `framework-espidf`
1. set in platformio.ini: `platform = https://github.com/atanisoft/platform-espressif32#espidf-arduino`
1. set in platformio.ini: `framework = arduino` and build (so that pio downloads the `framework-arduinoespressif32` package
1. set back to `framework = espidf` and build

atanisoft:
@kevinresol I'll try and reproduce it locally. This may require an update on the arduino-esp32 core though.

jakubtopic:
Hi, I wanted to get into this, but I'm certainly missing something. What do I have to do to make including of the Arduino libraries (like WiFi, Wire, ...) work? `#include <Arduino.h>` apparently works, but the compiler doesn't know where to find any "standard" Arduino library.

Do I need to explicitly tell the compiler where to look for these header files?

Example of error related to `#include "Preferences.h"`
```
In file included from src/ChargingControl.h:4:0,
from src/ChargingControl.cpp:5:
src/Config.h:11:25: fatal error: Preferences.h: No such file or directory
```

Content of my `platformio.ini`:
```
[env:esp32doit-devkit-v1]
platform = https://github.com/atanisoft/platform-espressif32#espidf-arduino
board = esp32doit-devkit-v1
framework = espidf
upload_port = /dev/ttyUSB0
monitor_speed = 115200
build_flags =
  -std=gnu++11
  -Wall
  -Wextra
  -Werror
  -D "FIRMWARE_VERSION=\"`git describe --tags --abbrev=0`\""
lib_deps =
  ...
```

atanisoft:
@jakubtopic did you list Preferences as an entry in lib_deps? That should be all that is required

jakubtopic:
@atanisoft No, I do not list any of these libraries. If I set `platform` to `arduino`, Library Dependency Finder is able to find all those libraries by itself. Explicit listing of 'Preferences' in lib_deps results in the same error.

jakubtopic:
I noticed @VCO34 did some tinkering with the build flags:
> -Ic: /Users/coufal/.platformio/packages/framework-arduinoespressif32@2.10000.0/libraries/Wire/src/
Is this really necessary?

atanisoft:
No that should not be necessary. Lib_deps should be used and working for this. I'll need to do some further testing and setup an example for it.

atanisoft:
Also build flags will only add headers but not add the sources of the library.

jakubtopic:
Ok, I figured out the problem with LDF... Sort of. It requires disabling compatibility mode (`lib_compat_mode = false`). In the current state, I can't compile some arduino libraries :slightly_frowning_face:. Like WiFiAP:
```
/home/topicja/.platformio/packages/framework-arduinoespressif32/libraries/WiFi/src/WiFiAP.cpp:40:43: fatal error: dhcpserver/dhcpserver_options.h: No such file or directory
compilation terminated.
*** [.pioenvs/esp32doit-devkit-v1/lib1f2/WiFi/WiFiAP.o] Error 1
In file included from /home/topicja/.platformio/packages/framework-arduinoespressif32/libraries/Update/src/Updater.cpp:5:0:
/home/topicja/.platformio/packages/framework-espidf/components/bootloader_support/include/esp_image_format.h:44:3: warning: 'esp_image_spi_freq_t' defined but not used [-Wunused-variable]
} esp_image_spi_freq_t;
^
```
I'm sorry, this is probably out of scope of this repository.

atanisoft:
That is very odd, it looks like something is not right in your IDF setup (or in the IDF PIO is using)

jakubtopic:
I don't know... At this moment I'm completely lost :sweat_smile: Could you please provide some example project? I know there is 'espidf-arduino-blink', but it does not use any Arduino libraries - just `Arduino.h`.

the-blest:
@kevinresol Still fairly new to all this, but I get the same error when building against esp-idf v3.1.3 (which I think is the latest stable in platformio). It seems like **esp_image_verify** is missing in that version, so there a mismatch between **esp_image_format.h** from esp-idf and arduino-esp32. Yesterday esp-idf v3.2 was released, which has the missing function. I haven't gotten it to build yet, but I hope it's fairly painless to upgrade.

kevinresol:
Any update please? Will this get fixed/merged?

sticilface:
it appears that @atanisoft has removed his github repo https://github.com/atanisoft/platform-espressif32#espidf-arduino. 

Any reason why?

atanisoft:
> it appears that @atanisoft has removed his github repo https://github.com/atanisoft/platform-espressif32#espidf-arduino.
> 
> Any reason why?

Primary reason is to reduce confusion. The cloned repo was very out of date and would lead to people filing issues which were due to where they sourced the platform-espressif32 bits from. I'll be recreating the repo and freshening this PR to the latest 1.9.0 platform-espressif32 once I have a few free cycles to focus on this.

sticilface:
ah ok.  this would be a great feature if @ivankravets would consider merging it.  I'm trying to get coredumps to work within the arduino-esp32 framework so i can debug an  easy to reproduce hard to debug crash.

atanisoft:
For those that still need this support please see the updated PR #193 which replaces this outdated version. The new PR also includes an example showing using the Arduino Libraries via lib_deps.

PR #179, opened 2019-05-11
Add BLE support, instead of removing it

mkx:
protocomm_ble should be part of the build if the two variables are set (see Makefiles for upstream espidf).

This has a potentially huge impact on firmware size, it might be necessary to change the default values in sdkconfig.h


valeros:
Thanks for the PR. Resolved in daa26141c6143b634378a52a4b24509a26103edf

PR #215, opened 2019-08-05
Remove "Automatically generated file; DO NOT EDIT." from "sdkconfig.h"

pfeerick:
Just noticed that this commit/change was lost in the a4c3a868ba18e2d1767ee40179d4d2aa03a4773b update to examples. 

Essentially re-applying 4646af1de26fa0eb8d5f2cf29856cb1b1c00f32a as well as updating subsequently added examples
`find ./examples/ -type f -name "sdkconfig.h" -exec sed -i -e "/* Automatically generated file; DO NOT EDIT./d" {} \;`

### DIFF
--- a/examples/espidf-arduino-blink/src/sdkconfig.h
+++ b/examples/espidf-arduino-blink/src/sdkconfig.h
@@ -1,6 +1,5 @@
 /*
  *
- * Automatically generated file; DO NOT EDIT.
  * Espressif IoT Development Framework Configuration
  *
  */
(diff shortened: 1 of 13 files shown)


dontsovcmc:
Wow! I was very surprised when found DO NOT EDIT in both files... 
This is helpful commit, thanks!

valeros:
Resolved in https://github.com/platformio/platform-espressif32/commit/0f7d2de7e2f15d0aa68469524c53d7bef47fadf0 Starting with ESP-IDF v4.0 `sdkconfig.h` is generated automatically in build directory

Issue #218, opened 2019-08-09
COMPONENT_EMBED_FILES not working in library.json


Dee-Fuse:
I'm trying to add binary data to the firmware as in #161 but bundled into an library.

I've added the build flag but no matter what I try to change it does not find / include the corresponding files.

Is it possible to include such a flag into a library or do I have to include it into the platformio.ini?

Dee-Fuse:
Its looks like the build options from an library.json file are not passed through `builder/frameworks/_embedtxt_files.py` to embedd files. The problem may lay in the platformio-core instead of the esp32 platform package.

valeros:
Hi @Dee-Fuse! Have you tried [extra_script](http://docs.platformio.org/en/latest/librarymanager/config.html#extrascript) option?

Dee-Fuse:
Yes, I've tryed it but without luck.
The problem with that option is, that the `extra_script` is been executed after `_emebdtxt_files.py`.
Due to that behaviour, the adding of the `CPPDEFINES` flags here never gets picked up by the embedding script.

I've tried:
```
Import('env')

env.Append(
    CPPDEFINES=["-D COMPONENT_EMBED_FILES www/style.css:www/jquery.gz:www/code.js:www/index.html"]
)
```
And 
```
global_env = DefaultEnvironment()
global_env.Append(
    CPPDEFINES=["-D COMPONENT_EMBED_FILES www/style.css:www/jquery.gz:www/code.js:www/index.html"]
)
```

Here is the output of the build process: ( I've added a print(env.get("CPPDEFINES")) in both files to verify it)
```
CONFIGURATION: https://docs.platformio.org/page/boards/espressif32/esp32dev.html
PLATFORM: Espressif 32 1.9.0 > Espressif ESP32 Dev Module
HARDWARE: ESP32 240MHz, 320KB RAM, 4MB Flash
DEBUG: Current (esp-prog) External (esp-prog, iot-bus-jtag, jlink, minimodule, olimex-arm-usb-ocd, olimex-arm-usb-ocd-h, olimex-arm-usb-tiny-h, olimex-jtag-tiny, tumpa)
PACKAGES: toolchain-xtensa32 2.50200.80 (5.2.0), framework-espidf 3.30202.190627 (3.2.2), tool-esptoolpy 1.20600.0 (2.6.0)
[...]

PR #220, opened 2019-08-20
Allow usage of COMPONENT_EMBED_FILES flags

### LINKED ISSUE #161: COMPONENT_EMBED_TXTFILES  modifies included files
There seems to be a bug where PlatformIO corrupts embedded files.

Following the directions here:  [http://docs.platformio.org/en/latest/platforms/espressif32.html#embedding-binary-data](http://docs.platformio.org/en/latest/platforms/espressif32.html#embedding-binary-data)

I created a test case.    Using PlatformIO to create a new project for the ESP32dev board, selecting the IDF (Not Arduino) env.

Create these three files:
**main.c** 
```
#include <stdint.h>


extern const uint8_t code_js_start[] asm("_binary_src_main_js_start");				
extern const uint8_t code_js_end[] asm("_binary_src_main_js_end");


void app_main() {
    
}
```

**main.js**
```
var mh = document.querySelector('h1');
mh.textContent = 'Hello World';
```

**platformio.ini**
```
; PlatformIO Project Configuration File
;
;   Build options: build flags, source filter
;   Upload options: custom upload port, speed and extra flags
;   Library options: dependencies, extra library storage
;   Advanced options: extra scripting
;
; Please visit documentation for the other options and examples
; https://docs.platformio.org/page/projectconf.html


[env:esp32dev]
platform = espressif32
board = esp32dev
framework = espidf
monitor_speed = 115200

build_flags =
    -DCOMPONENT_EMBED_TXTFILES=src/main.js

```

Before compiling the file size  of main.js is 70 bytes.   After compiling the file has been modified with an extra byte added to the end increasing the filesize of 71 bytes.  This extra byte is only added once, in that future 'compiles' will not continue to append an extra byte.  Looking at sample files which have been modified - it appears that a NULL byte(0x00)  is being appending to the end of text files before they are included.  And this NULL is carried forward into the binary image, as well as written back to the file on the disk.

This is very repeatable and seems to impact only text files (??).  Uploading a .gz (binary) file I do not notice this behavior.  (Though it could be the .gz already has a 0x00 ending)

Host:  Ubuntu 14.04lts
Visual Studio Code ver: 1.31.1
PlatformIO:  Home 2.0.0·Core 3.6.4

Dee-Fuse:
also copy original file bevore adding trailing zero when using `COMPONENT_EMBED_TXTFILES` to not leave it modified after embedding it

Fixes #161 and maybe #218

### DIFF
--- a/builder/frameworks/_embedtxt_files.py
+++ b/builder/frameworks/_embedtxt_files.py
@@ -14,6 +14,7 @@
 
 from os import SEEK_CUR, SEEK_END
 from os.path import basename, isfile, join
+import shutil
 
 from SCons.Script import Builder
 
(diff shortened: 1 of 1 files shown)


J-Rios:
Why is this not merged yet? It is really annoying that build process modify sources files (separate javascript files fails to be used if it is upload with any null terminator character, just first upload goes ok, after that, the developer must remove any null terminator character that has been added to the source file due last build).

valeros:
Many thanks to @Dee-Fuse ! Could you please retest with the latest platform from the development branch?

PR #223, opened 2019-09-08
Change default serial upload speed to 460k

pfeerick:
Whilst the esp8266/Arduino core has defaulted and limited their
upload speeds to 460800, the ESP32 core has not, but is subject
to the same issues, namely compatability issues with CH340 usb
serial chip and higher baud rates (possibly only on *nix based
systems).

There is possibly little benefit going higher than 460800 due
to flash erase and write times rather than serial transfer
times (even if it were to connect reliably).

sed -i "s/\"speed\".*/\"speed\": 460800/g" boards/*.json

### DIFF
--- a/boards/alksesp32.json
+++ b/boards/alksesp32.json
@@ -27,7 +27,7 @@
     "maximum_ram_size": 327680,
     "maximum_size": 4194304,
     "require_upload_port": true,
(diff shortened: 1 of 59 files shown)


valeros:
Thanks!

Issue #226, opened 2019-09-16
ESPIDF v3.3 requires "pyparsing"


atanisoft:
Using latest develop branch the following is observed while building:
```
generate_project_ld_script([".pio\build\pcb_oled_sh1106\esp32.project.ld"], ....... 
Traceback (most recent call last):
  File ".....platformio\packages\framework-espidf\tools\ldgen\ldgen.py", line 22, in <module>
    from fragments import FragmentFileModel
  File "......platformio\packages\framework-espidf\tools\ldgen\fragments.py", line 20, in <module>
    from sdkconfig import SDKConfig
  File ".....platformio\packages\framework-espidf\tools\ldgen\sdkconfig.py", line 18, in <module>
    from pyparsing import Word, printables, Combine, Literal, hexnums, quotedString, Optional, nums, removeQuotes, oneOf, Group, infixNotation, opAssoc
ImportError: No module named pyparsing
```
(above has been pruned to show only the required pieces)

After resolving the above by running "pip install pyparsing" in the penv/Scripts directory the following is observed:
```
IOError:.....platformio\packages\framework-espidf\Kconfig:59: Could not open '.......platformio\packages\framework-espidf\components\app_update\Kconfig.projbuild#......platformio\packages\framework-espidf\components\bootloader\Kconfig.projbuild#......platformio\packages\framework-espidf\components\esptool_py\Kconfig.projbuild#.....platformio\packages\framework-espidf\components\partition_table\Kconfig.projbuild' (ENOENT: No such file or directory). Perhaps the $srctree environment variable (which was unset) is set incorrectly. Note that the current value of $srctree is saved when the Kconfig instance
is created (for consistency and to cleanly separate instances). Also note that e.g. $FOO in a 'source' statement does not refer to the environment variable FOO, but rather to the Kconfig Symbol FOO (which would commonly have
'option env="FOO"' in its definition).
```

The above is due to Windows not defining IFS in the env and the default setting being " \t\n" (space, tab, newline) inside kconfiglib.py.

Using space instead of # resolves this issue and building is successful.

@valeros FYI...

valeros:
Thanks for reporting the problem! Strange, but our CI didn't catch that. Are you sure you're using the `esp-idf` package shipped with PlatformIO? The reason why I used `#` as the delimiter is because a lot of our users have a whitespace character is their username and `ldgen.py` simply doesn't take it into account.

atanisoft:
[...]

PR #227, opened 2019-09-16
Fix espidf delineation of kconfig args

atanisoft:
Use space instead of # to delineate kconfig args.

partial fix for https://github.com/platformio/platform-espressif32/issues/226

@valeros test output https://github.com/atanisoft/ESP32CommandStation/runs/224036767


valeros:
Thanks!

PR #240, opened 2019-10-21
[m5stick-c] Correct upload speed

1technophile:
The current upload speed gives an error when trying to upload on m5stick-c:
```
Configuring flash size...

A fatal error occurred: Timed out waiting for packet header
*** [upload] Error 2
```
With `1500000 `the upload works correctly.

Other reference; arduino IDE [m5stick-c definition](https://github.com/espressif/arduino-esp32/blob/c8d8dc226545a096a34bf5164551c087b658d33a/boards.txt#L2446)


valeros:
Thanks!

PR #249, opened 2019-11-22
Replace depreciated PROJECTSRC_DIR

pfeerick:
Per https://github.com/platformio/platformio-core/blob/44a9de6dcb51b748f36d19b0a5c77962c326fabb/platformio/builder/main.py#L109 this is a legacy variable


ivankravets:
Thanks!

PR #256, opened 2019-12-08
Correctly build with SPI RAM cache workaround

### LINKED ISSUE #254: PSRAM is unusable due to broken ESP-IDF builder
According to [documentation][1], adding some build flags is sufficient to enable PSRAM support and enable the PSRAM/SPI RAM cache workaround. For background, ESP32 chips have [a silicon bug][2] that causes memory inconsistency unless this workaround is enabled at compile & link time.

The documentation says to enable the workaround with the `-DCONFIG_SPIRAM_CACHE_WORKAROUND` build flag. However, this flag doesn't actually do anything. The flag itself is handled by the ESP-IDF build system, which PlatformIO doesn't use. So it has no actual effect.

Trying to use PSRAM results in memory inconsistencies and crashes. Random bytes in PSRAM will flip to zero and other weird stuff will happen. Overall, the system is completely unstable. This may also be an issue for Arduino users, but I've only tested with ESP-IDF.

To fix this problem, the build system needs to do a few different things when `CONFIG_SPIRAM_CACHE_WORKAROUND` is enabled:

1. Pass the `-mfix-esp32-psram-cache-issue` flag to the compiler to ensure only compatible code is emitted.
2. Ensure that `libc-psram-workaround.a` and `libm-psram-workaround.a` are linked instead of the default `libc.a` and `libm.a`.
3. Skip `esp32.rom.spiram_incompatible_fns.ld` during linking, since it is explicitly SPI RAM-incompatible. Instead, pass `esp32-spiram-rom-functions.lf` to `ldgen.py`.

Also, `ldgen.py` needs to be pointed to a valid `sdkconfig` file. Currently,  [`generate_project_ld_script`](https://github.com/platformio/platform-espressif32/blob/30f2d55355a1a7da37cb63b0778e5f8857eeacc1/builder/frameworks/espidf.py#L441) points it to `sdkconfig.h`, which it cannot process.

I've addressed these issues in a fork. I'm happy to open a pull request if this looks acceptable:
https://github.com/tidbyt/platform-espressif32/commit/69359a84e88c4d603ab202ef0f884f087ae93805

Finally, if you want PSRAM to work, you need these fixes and also need, at minimum, the following config in `sdkconfig`:

```
CONFIG_SPIRAM_SUPPORT=y
CONFIG_SPIRAM_CACHE_WORKAROUND=y

// either of the following will work
CONFIG_SPIRAM_USE_MALLOC=y
CONFIG_SPIRAM_USE_CAPS_ALLOC=y
```

I think #175 tries to address this, but I'm not sure if build flags are sufficient, since `ldgen` actually looks at the `sdkconfig` file directly.

[1]: https://docs.platformio.org/en/latest/platforms/espressif32.html#external-ram-psram
[2]: https://docs.espressif.com/projects/esp-idf/en/latest/api-guides/external-ram.html#esp32-rev-v1

rohansingh:
Fixes #254.


rohansingh:
@ivankravets @valeros Does this look acceptable to you?

valeros:
Hi @rohansingh ! Many thanks for the PR, looks great! Could you please also add a simple example to our CI that makes use of this feature?

rohansingh:
Alright, I've added an example that shows PSRAM usage. It seems to work, when I flash it to a chip I see this:

```
I (195) cpu_start: Starting scheduler on PRO CPU.
I (0) cpu_start: Starting scheduler on APP CPU.
I (206) spiram: Reserving pool of 32K of internal memory for DMA/internal allocations
Amount of free PSRAM: 4194264 bytes
```

Also, it looks like the correct libs are being linked. When using `platformio run -v`, I see `-lc-psram-workaround` and `-lm-psram-workaround`.

**Note:** this all only works if this line is added to `sdkconfig.h`:

```
#define CONFIG_SPIRAM_CACHE_WORKAROUND 1
```

Adding a build flag to `platformio.ini` is not sufficient, since `espidf.py` only looks at the `sdkconfig.h`, not the build flags. That's existing behavior that is outside the scope of this PR to change.

rohansingh:
Fixed merge conflict. Should be good to merge now.

rohansingh:
Looks like there are still some CI errors. I'll have to investigate at some point but I don't have the bandwidth at the moment.

davidjade:
@rohansingh I was trying to get this to work by merging it into my local config and I get linker errors. I think I found the problem though. This new version of espidf.py seems to have an error in it, line 455.

Currently it is this:
```python
["\"%s\""` % f for f in project_files.get("lf_files")]),
```
but a few lines above this there is this new code added but never used:
```python
lf_files = project_files.get("lf_files") + env["LD_FRAGMENTS"]
```
So I think line 455 should be this:
```python
["\"%s\"" % f for f in lf_files]),
```
Otherwise env["LD_FRAGMENTS"] never has a way to get into the build. 

I've changed this in my local config and it fixes things (and makes PSRAM work, at least according to the free heap size).

rohansingh:
@davidjade Wow, great catch. I just checked and our local fork at [tidbyt/platform-espressif32](https://github.com/tidbyt/platform-espressif32) actually has the exact fix that you described, but it just never made it into this PR branch. I'll pull it in now.

rohansingh:
OK, looks like that line got lost while fixing merge conflicts. Unfortunately, `espidf.py` has changed a ton now so this PR mostly has to be redone.

rohansingh:
Yup, looks like #297 rewrote `espdif.py` for ESP-IDF v4.0. It looks like it now actually runs `cmake`, so maybe this is resolved by coincidence?

Either way, this PR is dead in the water.

PR #258, opened 2019-12-16
Update ttgo-lora32-v1.json

donnacha:
Added link to the actual vendor github page for board, rather than just a google search link.


valeros:
Thanks!

PR #290, opened 2020-02-26
Add ELF file hash to images generated by elf2image

### LINKED ISSUE #272: Firmware binaries are missing SHA256 hash of the original ELF
When building binaries from an ELF, the ESP-IDF build system adds a SHA256 hash of the original ELF at offset `0xb0`. This is vital for setting up ESP OTA correctly, since the SHA256 hash is an important method of checking whether the device needs to update its firmware, whether an update succeeded, etc.

In pratical terms, this means that [`esp_ota_get_app_description`][1] always returns an `esp_app_desc_t*` with an empty SHA256 hash, and [`esp_ota_get_app_elf_sha256`][2] returns an empty array as well.

You can see the following logs when an ESP-IDF app built with PlatformIO boots up:

```
I (742) cpu_start: Application information:
I (747) cpu_start: Project name:     firmware
I (752) cpu_start: App version:      1.0.0
I (757) cpu_start: Compile time:     Jan 11 2020 01:47:20
I (763) cpu_start: ELF file SHA256:  0000000000000000...
I (769) cpu_start: ESP-IDF:          aafab7d68
```

The `0000000000000000` should contain the SHA256 prefix of the ELF file.

[1]: https://docs.espressif.com/projects/esp-idf/en/v3.3/api-reference/system/ota.html#_CPPv427esp_ota_get_app_descriptionv
[2]: https://docs.espressif.com/projects/esp-idf/en/v3.3/api-reference/system/ota.html#_CPPv426esp_ota_get_app_elf_sha256Pc6size_t

rohansingh:
Fixes #272.

### DIFF
--- a/builder/main.py
+++ b/builder/main.py
@@ -180,6 +180,7 @@ env.Append(
                 "--flash_mode", "$BOARD_FLASH_MODE",
                 "--flash_freq", "${__get_board_f_flash(__env__)}",
                 "--flash_size", board.get("upload.flash_size", "detect"),
+                "--elf-sha256-offset", "0xb0",
                 "-o", "$TARGET", "$SOURCES"
             ]), "Building $TARGET"),
             suffix=".bin"


valeros:
Many thanks!

PR #291, opened 2020-02-27
Add erase_otadata target

rohansingh:
This new target wipes out any OTA data partition, ensuring that the
device boots from the factory partition.


ivankravets:
It looks we should implement https://github.com/platformio/platformio-core/issues/1889

See workaround https://docs.platformio.org/en/latest/projectconf/advanced_scripting.html#custom-target

PR #320, opened 2020-04-14
allow user to change spiffs.bin file name via pre:script

giobauermeister:
I noticed that is possible to change default PROGNAME 'firmware.bin' filename via pre:script, but File System Image filename 'spiffs.bin' is hardcoded. I added a new FSIMAGENAME variable that can be modified via pre:script as well.


valeros:
Thanks, resolved in 856885e2196c2acda4bf4c36b010b427db38747c

PR #321, opened 2020-04-15
JTAG debug support for mhetesp32minikit board with ftdi.

rafacouto:


valeros:
Hi @rafacouto ! You mentioned that this board has an on-board debug probe based on FTDI chip, but I can only see UART-USB bridge on the board. Could you please clarify?

rafacouto:
> Hi @rafacouto ! You mentioned that this board has an on-board debug probe based on FTDI chip, but I can only see UART-USB bridge on the board. Could you please clarify?

It was an error: no ftdi on-board. I have tested it with ftdi2232hl:

mhetesp32minikit ft2232hl

I've sent the patch right now. Thank you for the comment.

PR #325, opened 2020-04-19
Add SVD for ESP32 (wrover kit)

jenswet:
See https://community.platformio.org/t/clion-debugging/12534/2


valeros:
Hi @jenswet ! Thanks for this PR. Strange, it looks like there is an issue with SVD parser in VSCode:
```
Unable to parse SVD file: TypeError: Cannot read property '0' of undefined
```
In the meantime, could you please also add this file to other boards? We'll merge it when the issue with VSCode is resolved.

Thanks!

jenswet:
Hi @valeros I added it to all boards.
Could you verify that CLion can parse it on your end?
I could reproduce the error in VSCode. Unfortunately I cannot guarantee that my SVD is valid, because I am not really into this topic. But at least the XML is valid according to an online validator tool.
Regards!

ivankravets:
@jenswet Thanks for the PR! VSCode can't parse this file https://github.com/platformio/platformio-vscode-ide/issues/1675

Maybe, this is our issue.

Could you try to experiment with VSCode and removing some parts from SVD to find an issue? For example, start from something simple and add more and more lines. You will find where an issue arises.

jenswet:
Is there some more verbose build of the vscode plugin available?

It would be much easier to find the source of the error, when there is more logging like the line number of the parsing error or other information.

ivankravets:
Does it work now with VSCode?

maxgerhardt:
I've also noted a similiar parsing error in https://github.com/platformio/platformio-vscode-ide/issues/2368 and CLion can read this file though. Would be really nice to get attention to this issue and get a SVD file for the ESP32 (or a fixed parser) for peripheral register viewing! :)

PR #350, opened 2020-05-20
Use latest esptool.py 2.8

Jason2866:


valeros:
Do you have any issues with 2.6.x? Looks like Arduino framework is still using it.

Jason2866:
Yes, had issues with ESP32 flash encryption. Solved with v.2.8

Jason2866:
Closing since 3.0 is now used

PR #378, opened 2020-07-14
Change define "ARDUINO_T-Watch" to "ARDUINO_T_Watch" prevent "warning: ISO C99 requires whitespace after the macro name"

sweetlilmre:
Change define "ARDUINO_T-Watch" to "ARDUINO_T_Watch" prevent "warning: ISO C99 requires whitespace after the macro name"

### DIFF
--- a/boards/ttgo-t-watch.json
+++ b/boards/ttgo-t-watch.json
@@ -4,7 +4,7 @@
       "ldscript": "esp32_out.ld"
     },
     "core": "esp32",
-    "extra_flags": "-DARDUINO_T-Watch -DBOARD_HAS_PSRAM -mfix-esp32-psram-cache-issue",
+    "extra_flags": "-DARDUINO_T_Watch -DBOARD_HAS_PSRAM -mfix-esp32-psram-cache-issue",
     "f_cpu": "240000000L",
     "f_flash": "40000000L",
     "flash_mode": "dio",


sweetlilmre:
Any chance of this pull request getting some attention? It's a very small fix with little to zero impact and cuts down compiler noise significantly.

valeros:
Hi @sweetlilmre ! Sorry for the late reply. How is this board defined in Arduino IDE? You can always suppress any unwanted flags using `build_unflags` option.

sweetlilmre:
Hi @valeros 

As far as I can see this definition is platformio specific and not necessarily an unwanted flag. The other TTGO examples have underscores as opposed to dashes in the extra_flags definitions (preventing this warning) and the definition is useful for gating watch specific functionality, although currently unused.
I'm assuming this was just a typo in the original submission?

valeros:
I checked this board in Arduino IDE and the definition is the same `-DARDUINO_T-Watch`. We tend to use the same board configuration just to make sure users won't bump into any problems caused by different definitions.

sweetlilmre:
Hi @valeros 

I've opened a pull request against the espressif repo to change the ttgo-t-watch.build.board=T-Watch definition.
Let's see if we can get it fixed there first.

sweetlilmre:
espressif has fixed this issue by changing the definition to "TWatch" in this commit: https://github.com/espressif/arduino-esp32/commit/6f237a84152d59537a012e7bfc3edbabd6adbd9d#diff-fce09ad94aad2ba67472cd9142f00146

@valeros how would you like to proceed? should I close this and resubmit a patch or will you fix it on your side for consistency?

valeros:
Thanks for the reminder, the define will be fixed  as soon as the Espressif team publishes a new release.

sweetlilmre:
espressif has fixed this issue by changing the definition to "TWatch" in this commit: https://github.com/espressif/arduino-esp32/commit/6f237a84152d59537a012e7bfc3edbabd6adbd9d#diff-fce09ad94aad2ba67472cd9142f00146

@valeros how would you like to proceed? should I close this and resubmit a patch or will you fix it on your side for consistency?

PR #420, opened 2020-10-20
add M5Atom AND Core2 Board

Gitshaoxiang:


Gitshaoxiang:
Board is now available also in Arduino-Framework: http://github.com/espressif/arduino-esp32 so ready for merging

Issue #177, opened 2019-05-01
Bug uploadfs with tool-openocd-esp32


antoniotorres:
Cannot upload spiffs image with `upload_protocol = esp-prog` through JTAG

👇 Offset is hardcoded to factory image region. 
https://github.com/platformio/platform-espressif32/blob/edc118b22aced53d45d19fe231f7dca4368db88f/builder/main.py#L321

I'm using a custom partition table and load correctly.
It does work with `upload_protocol = esptool` through UART

antoniotorres:
Example of `platformio.ini`. Using esp32-wroom-32d (16Mb)
```
[env:ftdi16mb]
platform = espressif32@1.8.0
board = esp32doit-devkit-v1
framework = espidf
upload_protocol = esp-prog
upload_speed = 115200
monitor_speed = 115200
board_build.partitions = custom_partition_table.csv
board_upload.flash_size = 16MB
build_flags = -DCOMPONENT_EMBED_TXTFILES=src/certs/aws_s3_ca.pem -DCOMPONENT_EMBED_TXTFILES=src/certs/aws_root_ca.pem -Wl,-Map=output.map
extra_scripts = pre:scripts/pre-compile.py
```
Example of `custom_partition_table.csv`
```
# Name,   Type, SubType, Offset,   Size
# Note: if you change the phy_init or app partition offset, make sure to change the offset in Kconfig.projbuild
nvs,      data, nvs,     ,         0x4000
otadata,  data, ota,     ,         0x2000
phy_init, data, phy,     ,         0x1000
factory,  0,    0,       ,         1M
coredump, data, coredump,,         64K
ota_0,    0,    ota_0,   ,         1M
ota_1,    0,    ota_1,   ,         1M
spiffs,  data, spiffs,   ,         1M
```

antoniotorres:
Command sequence
`pio run -e ftdi16mb -t erase --upload-port /dev/cu.SLAB_USBtoUART -v` (erase only works with uart)
`pio run -e ftdi16mb -t upload -v` (upload firmware jtag)
`pio run -e uart16mb -t uploadfs --upload-port /dev/cu.SLAB_USBtoUART -v` (upload spiffs uart)

PR #422, opened 2020-10-24
Fix uploadfs through JTAG

morgendagen:
This pull request fixes https://github.com/platformio/platform-espressif32/issues/177.


meltdown03:
You should change `program_esp32` to `program_esp` in your PR. That is the new command for flashing.

PR #424, opened 2020-10-28
added option to override fw app offset when customizing partition table

ramidomalys:
added option to override fw app offset when customizing partition table offset ( was always fixed to 0x10000 in esptool upload methode ) -> changed to adapt to option :  board_upload.offset_address= x


valeros:
Thanks, fixed in https://github.com/platformio/platform-espressif32/commit/115f98151ed2e8328ee392c62752bd4678b8b5e5

PR #436, opened 2020-11-25
Replace '-' in macro name with '_' for board "TTGO T-Watch"

treibholz:
Replace '-' in macro name with '_' for board "TTGO T-Watch"

Fixes compiler warnings:

    <command-line>:0:10: warning: ISO C++11 requires whitespace after the macro name

This is similar to #81

### DIFF
--- a/boards/ttgo-t-watch.json
+++ b/boards/ttgo-t-watch.json
@@ -4,7 +4,7 @@
       "ldscript": "esp32_out.ld"
     },
     "core": "esp32",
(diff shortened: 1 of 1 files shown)


valeros:
Hi @treibholz ! Thanks for the PR. Looks like it's already fixed in the framework here https://github.com/espressif/arduino-esp32/pull/4212 . So the variant will be updated accordingly when a new version of the framework is released.

Issue #459, opened 2020-12-28
Make adapter_speed changable in platformio.ini


maxgerhardt:
Many uploading / debugging problems over JTAG with the ESP32 are due to a too high JTAG speed. 

In previous versions of platform-espressif32, one could just change the hardcoded `adapter_khz` value in the `.cfg` configuration files of the `tool-openocd` package. However, this is not possible anymore, since now the platform issues an explicit `-c adapter_khz <speed>` command.

https://github.com/platformio/platform-espressif32/blob/a58a358fdc1122523c7fcf7b4fc8b4016e48961d/platform.py#L85-L86

https://github.com/platformio/platform-espressif32/blob/a58a358fdc1122523c7fcf7b4fc8b4016e48961d/platform.py#L124-L129

When using e.g. the `platformio.ini` 

```
[env:esp32dev]
platform = espressif32
board = esp32dev
framework = arduino
upload_protocol = esp-prog
```
then a "verbose upload" shows the invocation

>openocd -d2 -s C:/Users/Max/.platformio/packages/tool-openocd-esp32/share/openocd/scripts -f interface/ftdi/esp32_devkitj_v1.cfg -f board/esp-wroom-32.cfg -c "adapter_khz 20000" -c "program_esp {{...

But when attempting to change it via adding 

```
board_debug.adapter_speed = 100
```

The invocation does not change and still uses the 20000kHz speed. 

I think the problem is that it always gets this value from the board's manifest file, and dynmic changes to it via `board_<section>.<attribute> = <some value>` are ignored. 

I think it is very critical that this value can be changed without altering config files or the board's JSON manifest. Can the code be adapted so that dynmic changes are accepted? 

Also see topic https://community.platformio.org/t/debugging-problems/18059/3.

PR #463, opened 2021-01-02
Use a "safe" value for OpenOCD adapter speed

ryanplusplus:
@maxgerhardt opened an issue to make the adapter speed configurable via `platformio.ini` (see #459), but until that's possible, I think it would be best to set a "safe" speed by default.


valeros:
Thanks for the PR, resolved in https://github.com/platformio/platform-espressif32/commit/7e5f3f56fada130e7b7bb4b29330ed45b72af626

PR #469, opened 2021-01-19
Fix wrong macro definition of TWATCH

lewisxhe:


valeros:
Thanks, duplicate https://github.com/platformio/platform-espressif32/pull/436 . The variant will be updated together with a new Arduino core release.

PR #493, opened 2021-02-16
Add Lolin32 Lite and TTGO T7 boards

balrog-kun:
Add Wemos Lolin32 Lite (original no longer being produced, same as Lolin32), TTGO
T7 v1.3 and v1.4 boards (status unknown).  All three are rather generic ESP32 boards the
main difference being pin layout.

I use these from ESPHome which relies on the platformio board definitions.  On top of it ESPHome has a per-board pin assignemtn registry (status LED pin, VBAT pin, etc.) where these boards actually differ.

### DIFF
--- a/boards/lolin32-lite.json
+++ b/boards/lolin32-lite.json
@@ -1,0 +1,34 @@
+{
+  "build": {
+    "arduino":{
+      "ldscript": "esp32_out.ld"
+    },
+    "core": "esp32",
+    "extra_flags": "-DARDUINO_LOLIN32_Lite",
+    "f_cpu": "240000000L",
+    "f_flash": "40000000L",
+    "flash_mode": "dio",
+    "mcu": "esp32",
+    "variant": "lolin32"
+  },
+  "connectivity": [
+    "wifi",
+    "bluetooth",
+    "ethernet",
+    "can"
+  ],
+  "frameworks": [
+    "arduino",
+    "espidf"
+  ],
+  "name": "WEMOS LOLIN32 Lite",
+  "upload": {
+    "flash_size": "4MB",
+    "maximum_ram_size": 532480,
+    "maximum_size": 4194304,
+    "require_upload_port": true,
+    "speed": 460800
+  },
+  "url": "https://web.archive.org/web/20191002041532/https://wiki.wemos.cc/products:lolin32:lolin32_lite",
+  "vendor": "WEMOS"
+}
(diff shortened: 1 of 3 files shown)


valeros:
Hi @balrog-kun !

Thanks for the PR. Users often complain about PlatformIO when they upload a simple blink example and it doesn't work because of the messed up pin layout. That's why we tend to add boards that have officially supported variants. The `lolin32-lite` board seems fine because you mentioned that it's the same as  `Lolin32`, so I guess it has the same pin layout. As for the `TTGO T7` boards, you specified `ttgo-t7-v13-mini32` as the variant and I don't see anything close to that in the [Arduino core for ESP32](https://github.com/espressif/arduino-esp32/tree/master/variants).

balrog-kun:
> Thanks for the PR. Users often complain about PlatformIO when they upload a simple blink example and it doesn't work because of the messed up pin layout. That's why we tend to add boards that have officially supported variants. The `lolin32-lite` board seems fine because you mentioned that it's the same as `Lolin32`, so I guess it has the same pin layout.

Not really, it's a cut-down version and has fewer pins and the LED is on a different pin.

> As for the `TTGO T7` boards, you specified `ttgo-t7-v13-mini32` as the variant and I don't see anything close to that in the [Arduino core for ESP32](https://github.com/espressif/arduino-esp32/tree/master/variants).

Right, neither of the boards has arduino support.  I added them locally in `platform-espressif32`, `framework-arduinoespressif32` and `esphome` to be able to build for them, but I noticed that things would compile correctly even without the changes in `framework-arduinoespressif32`.  So I wasn't sure what order to send pull requests in -- do you reckon I should first send a pull-request to arduino-esp32?

balrog-kun:
Oh, by saying 

> (original no longer being produced, same as Lolin32)

I meant that both have been retired by the original manufacturer (wemos) and only clones continue to be available.

valeros:
> Not really, it's a cut-down version and has fewer pins and the LED is on a different pin.

That's exactly what I was trying to say. Imagine how frustrating it is to upload a simple sketch that uses `LED_BUILTIN` and it doesn't work.

> So I wasn't sure what order to send pull requests in -- do you reckon I should first send a pull-request to arduino-esp32?

Sounds like a good idea, Arduino IDE users will also benefit from your PR. Once they merge it and release a new stable version, I'll merge your PR here.

balrog-kun:
Done, so resuming this is now blocked by https://github.com/espressif/arduino-esp32/pull/4819 and a subsequent arduino-esp32 release.

PR #582, opened 2021-07-09
Esp32ExceptionDecoder: support uppercase hex addresses in backtrace

Tasssadar:
In some ESP-IDF version, after calling `esp_backtrace_print`, the backtrace is in uppercase.


valeros:
Many thanks!

PR #590, opened 2021-07-23
boards: add arduino support to ESP32-C3-DevKitM-1

stintel:
Without this, the board cannot be used in ESPHome.


Jason2866:
It is there in the branch https://github.com/platformio/platform-espressif32/tree/feature/arduino-idf-master

valeros:
It will be available once a stable version of Arduino for ESP32 is released.

PR #601, opened 2021-08-11
Updating the espidf-aws-iot example to use AWS Embedded C Embedded SDK v3.1.5

ajlennon:
I am trying to support OTA which appears to require that I use a recent embedded SDK (e.g. 202103.00)

I am finding really hard going trying to work out how to move from the v3.0.1 e-SDK with ESP32 port here to the 202103.00 e-SDK as there has been quite a lot of refactoring.

There is a migration document from 3.x.x to 202103.00 but it is isn't making sense to me at the moment.

So as an intermediate step I have updated to the latest e-SDK v3.x.x series release which is v3.1.5

There were some changes and I've modified the pub/sub example as needed.

I've tested this on an M5Stack Grey here against my AWS IoT endpoint and am seeing pubs echoed back as subs so all is working.

Cheers, Alex


examples/espidf-aws-iot/platformio.ini
```diff
@@ -12,6 +12,7 @@ platform = espressif32
 framework = espidf
 board = esp32dev
 monitor_speed = 115200
+monitor_filters = time, colorize
```
valeros, line 15: Looks like this setting is project specific. Do you mind removing it?

valeros (changes requested):

ajlennon:
No it's not project specific. The demo as it stands outputs a load of crap making it very hard to see what's going on.

This fixes the demo.

examples/espidf-aws-iot/platformio.ini
ajlennon: I misundersood your point. I feel this is necessary as it correctly interprets the escape codes that the demonstration project outputs. Otherwise the codes make it hard to see what is going on. However it's separate from the E-SDK update so will break this out into another PR yes

ajlennon:
OK the change to the project to fix the parsing of the ESP32 output is removed - I will submit a separate PR

ajlennon:
NB. I would _really_ appreciate some advice on how to migrate from v3.1.5 to 20210301 as I am finding it really opaque!

ajlennon:
HI @valeros is this ready to go in?

valeros:
Many thanks, merged.

ajlennon:
Thanks @valeros !

PR #603, opened 2021-08-12
Fix AWS IoT demo output parsing

ajlennon:
The demonstration builds and runs on the ESP32 outputting
escape codes which are not parsed by default. This makes
the output hard to read.

This PR sets platform.io to correctly parse the escape codes.


valeros:
Thanks, merged.

PR #612, opened 2021-08-26
esp32_exception_decoder: ESP32-C3 support; possibility for external configuration

ivmarkov:
Dear all,

This PR is implementing the following set of changes to the `filter_exception_decoder.py` (a.k.a. `esp32_exception_decoder` filter):
- (Most important) Support for ESP32-C3 (and future RiscV based chips)
  - The stacktraces generated by the ESP RiscV chip family do not really start with a `Backtrace: ` string prefix (these start with a `Stack memory:\n\n` string prefix instead and overall have a very different and much more verbose dump of hexadecimal integers)
  - Rather than trying to extract the PC counters from this set of hex numbers, I've switched the filter to operate in a way similar to how the standard ESP IDF monitor operates (references embedded in the code itself): The ESP IDF monitor scans for *any* hex numbers that start with `0x4xxxxxxx` instead, as this is an indication of a memory address which is corresponding to a code location. While this results in hexadecimals outside of the stacktrace being decoded as program locations, it is (a) more robust and (b) it has the nice benefit *precisely* that PC counters anywhere in the log will be decoded. I.e. we are planning to dump backtraces on non-panic situations from our Rust code, so this type of decoding will decode these backtraces too

- As a consequence, the decoded output is served on a separate line (or lines) following the line where PC counter address (or addresses) were detected. Looks quite OK IMO

- The decoded output is colored (with an ANSI sequence) in yellow, for better visibility. ESP-IDF logging uses ANSI sequences by default too
- Finally (but important for us) I've put provisions in the code to to take the firmware location, `addr2line` location and the project location via environment variables, as this is necessary for our Rust support (PlatformIO is used there as a build-support utility, but is not driving the whole build process, which is orchestrated by Rust Cargo). If these environment variables are not set, the filter will fall back to the standard logic of locating these items based on PlatformIO APIs.

### DIFF
--- a/monitor/filter_exception_decoder.py
+++ b/monitor/filter_exception_decoder.py
@@ -31,53 +31,74 @@ class Esp32ExceptionDecoder(DeviceMonitorFilter):
 
     def __call__(self):
         self.buffer = ""
-        self.backtrace_re = re.compile(
-            r"^Backtrace: ?((0x[0-9a-fA-F]+:0x[0-9a-fA-F]+ ?)+)\s*"
-        )
+        # regex matches potential PC value (0x4xxxxxxx)
+        # Logic identical to https://github.com/espressif/esp-idf/blob/master/tools/idf_monitor_base/constants.py#L56
+        self.pcaddr_re = re.compile(r'0x4[0-9a-f]{7}', re.IGNORECASE)
 
         self.firmware_path = None
         self.addr2line_path = None
         self.enabled = self.setup_paths()
 
-        if self.config.get("env:" + self.environment, "build_type") != "debug":
-            print(
+        return self
+
+    def setup_paths(self):
+        self.project_dir = path_to_unicode(os.path.abspath(self.project_dir))
+
+        self.project_strip_dir = os.environ.get("esp32_exception_decoder_project_strip_dir")
+        self.firmware_path = os.environ.get("esp32_exception_decoder_firmware_path")
+        self.addr2line_path = os.environ.get("esp32_exception_decoder_addr2line_path")
+
+        if self.project_strip_dir is None:
+            self.project_strip_dir = self.project_dir
+
+        try:
+            if self.firmware_path is None or self.addr2line_path is None:
+                # Only load if necessary, as the call is expensive
+                data = load_project_ide_data(self.project_dir, self.environment)
+
+            if self.firmware_path is None:
+                # Only do this check when the firmware path is not externally provided
+                if self.config.get("env:" + self.environment, "build_type") != "debug":
+                    print(
                 """
 Please build project in debug configuration to get more details about an exception.
 See https://docs.platformio.org/page/projectconf/build_configurations.html
 
 """
-            )
+                    )
+                self.firmware_path = data["prog_path"]
 
-        return self
-
-    def setup_paths(self):
-        self.project_dir = path_to_unicode(os.path.abspath(self.project_dir))
-        try:
-            data = load_project_ide_data(self.project_dir, self.environment)
-            self.firmware_path = data["prog_path"]
             if not os.path.isfile(self.firmware_path):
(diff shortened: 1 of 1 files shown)


agners:
@ivmarkov FWIW, I noticed the missing backtrace on ESP32-C3/RISC-V too when adding support to ESPHome a while back:
https://github.com/esphome/esphome/pull/2062#issuecomment-886501461

You can "safely" decode the PC and LR to get at least the current function and the function before that (minus inlines, of course). Maybe this is  start.
https://github.com/esphome/esphome/pull/2062/files#diff-dbf001f542fb85de707889ca75fa11637d2f7cbd08c826b8abe15ac512789c7d

I think this is a limitation of the toolchain and/or ESP-IDF's exception handling which cannot make sense of the stack currently and hence print a useable Backtrace.

ivmarkov:
> I think this is a limitation of the toolchain and/or ESP-IDF's exception handling which cannot make sense of the stack currently and hence print a useable Backtrace.

It is a RiscV-specific limitation in that the code on the ESP32-C3 is compiled without FP (frame pointer support), and therefore the stack trace cannot be reconstructed.

- Whether it would be possible to enable FPs in GCC (and four our case - in rustc as well), is yet to be seen, but that would be one avenue of exploration.
- Another option is to enable eh_frame support (DWARF tables) inside the ESP-IDF menuconfig with `CONFIG_ESP_SYSTEM_USE_EH_FRAME `. This will get us proper backtraces, but what I'm afraid of is that these tables include symbol (function) names so this option might be prohibitively expensive. Yet still worth exploring because **some** of this info _must_ somehow be present even when `CONFIG_ESP_SYSTEM_USE_EH_FRAME` is disabled, because the C++ stack unwinding needs it anyway
- A third option is to just accept the "let's print the stack memory on RiscV instead of backtrace" status quo and try to decode it in the monitor using GDB, just like Espressif's own monitor does it, via GDB

More info here: https://github.com/esp-rs/rust/issues/76#issuecomment-909029830

monitor/filter_exception_decoder.py
```diff
@@ -31,53 +31,74 @@ class Esp32ExceptionDecoder(DeviceMonitorFilter):
...
-        if self.config.get("env:" + self.environment, "build_type") != "debug":
-            print(
+        return self
+
+    def setup_paths(self):
+        self.project_dir = path_to_unicode(os.path.abspath(self.project_dir))
+
+        self.project_strip_dir = os.environ.get("esp32_exception_decoder_project_strip_dir")
+        self.firmware_path = os.environ.get("esp32_exception_decoder_firmware_path")
+        self.addr2line_path = os.environ.get("esp32_exception_decoder_addr2line_path")
```
ivankravets, line 49: What do mean these hacks?

ivankravets:
@Tasssadar, sorry that disturb you. Does Espressif have an official decoder which can reuse without reinventing the wheels?

ivmarkov:
It had been more than a year, and in the meantime things have changed a bit:
- Regarding support for Rust: this is no longer necessary. Rust ESP toolchain now has its own monitor and exception backtrace decoder (albeit for xtensa only, but if necessary, it can be extended for riscv as well). So this aspect of the patch (the "hacks " as you say) are no longer necessary
-

pat1:
in documentation https://docs.platformio.org/en/latest/core/userguide/device/cmd_monitor.html#filters esp32_exception_decoder  I found  "Custom filter for [Espressif 32](https://docs.platformio.org/en/latest/platforms/espressif32.html#platform-espressif32) which decodes crash exception"  so I understand it works  for ESP32 -C3 too.
But it is not true.

bmedici:
I can confirm it does not wirk with ESP32C3 for me, neither.

bmedici:
Up ?

PR #689, opened 2021-12-18
add board Wireless-Tag WT32-ETH01 (wt32-eth01)

mondbaron:
Added board definition for WT32-ETH01 from Wireless-Tag (see http://www.wireless-tag.com/portfolio/wt32-eth01/)

- the corresponding variant `wt32-eth01` was recently added to https://github.com/espressif/arduino-esp32
- the variant includes the pin definitions for Ethernet PHY (LAN8720A)
- tested: works out-of-the-box with AsyncTCP and ESPAsyncWebServer

### DIFF
--- a/boards/wt32-eth01.json
+++ b/boards/wt32-eth01.json
@@ -1,43 +1,39 @@
 {
-    "build": {
-        "arduino": {
-            "ldscript": "esp32_out.ld"
-        },
-        "core": "esp32",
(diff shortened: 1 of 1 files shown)


mondbaron:
The corresponding variant has meanwhile been officially released with arduino-esp32 v2.0.2, so this board definition should work using the latest stable version.

PR #693, opened 2021-12-23
Toward ESP-IDF 5.0 support

X-Ryl669:
Hi,

  I've done some work toward supporting the latest ESP-IDF (4.4 and up).
I've fixed the changes in ESP-IDF file structure and linker directives.

~~Right now, everything builds fine except for the final linker stage where I'm getting linker relocations errors I don't understand~~.

EDIT: Should be fixed now.

### DIFF
--- a/builder/frameworks/espidf.py
+++ b/builder/frameworks/espidf.py
@@ -341,7 +341,7 @@ def extract_link_args(target_config):
             if fragment.startswith("-l"):
                 link_args["LIBS"].extend(args)
             elif fragment.startswith("-L"):
-                lib_path = fragment.replace("-L", "").strip()
+                lib_path = fragment.replace("-L", "").strip().strip('"')
                 _add_to_libpath(lib_path, link_args)
             elif fragment.startswith("-") and not fragment.startswith("-l"):
                 # CMake mistakenly marks LINKFLAGS as libraries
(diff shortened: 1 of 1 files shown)


X-Ryl669:
Ok, I've found the issue and fixed it. It should be working now. Please review...

jonathandreyer:
Hi @X-Ryl669,
Thanks for amazing job to towards support on upcoming versions of ESP-IDF. I have started to support the ESP32-S3 on PlatformIO and I was stuck without your contribution.
For my side, my goal is to use the upcoming version (v4.4) to have the support of ESP32-S3. During the integration of your contribution, I have noticed that the *ldgen* tool of v4.4 doesn't support the *--fragments-list*. I have also the feeling that the * section_input* args is not necessary. Can you confirm that?

I will publish it here when this will be a bit less dirty.

Best regards,

Jonathan Dreyer

X-Ryl669:
Honestly, I've no idea about what is required or not, I've just captured what idf.py is doing and modified the platformio script until it produced the same arguments. Concerning `--fragments-list`, there are 2 ways to pass the fragments to the script. platformio used to use fragments list, but idf.py does not. The ways arguments are built is different (a big string for the former, multiple args for the latter). 

Shouldn't be hard to change, but it needs to be done ;-).

jonathandreyer:
Thanks. I have created a new branch ([here](https://github.com/jonathandreyer/platform-espressif32/tree/espidf-440)) with those modifications and few cleanings.

X-Ryl669:
Feel free to trigger a new PR and I'll close this one, since yours is further away. Is there any maintainer around here anyway ?

brainstorm:
Duplicate of #711 and #716?... observing the same Saola-1 breakage described in https://github.com/platformio/platform-espressif32/pull/716#issuecomment-1079499068 though :/

PR #702, opened 2022-01-10
Use gcc-ar and gcc-ranlib

krzmaz:
In order to use link time optimizations (lto) `ar` and `ranlib` need to have the lto plugin supplied. This is achieved easiest by using the `gcc-*` equivalents. More info:
https://embeddedartistry.com/blog/2020/04/13/prefer-gcc-ar-to-ar-in-your-buildsystems/
This won't affect most builds at all, because lto [is disabled in the framework package](https://github.com/espressif/arduino-esp32/blob/master/tools/platformio-build-esp32.py#L83). As far as I've found this setting comes from https://github.com/espressif/esp-idf/issues/3989 and is more of a hotfix for Mac builds than a needed change.

I've successfully enabled lto in [my project](https://github.com/TheRealMazur/LinakDeskEsp32Controller/commit/79d6403a0d6f9f4d41c015a49761c870867d9181) without any functional differences, achieving ~5% flash usage reduction.

### DIFF
--- a/builder/main.py
+++ b/builder/main.py
@@ -146,14 +146,14 @@ env.Replace(
     __get_board_f_flash=_get_board_f_flash,
     __get_board_flash_mode=_get_board_flash_mode,
 
-    AR="%s-elf-ar" % toolchain_arch,
+    AR="%s-elf-gcc-ar" % toolchain_arch,
     AS="%s-elf-as" % toolchain_arch,
     CC="%s-elf-gcc" % toolchain_arch,
     CXX="%s-elf-g++" % toolchain_arch,
     GDB="%s-elf-gdb" % toolchain_arch,
     OBJCOPY=join(
         platform.get_package_dir("tool-esptoolpy") or "", "esptool.py"),
-    RANLIB="%s-elf-ranlib" % toolchain_arch,
+    RANLIB="%s-elf-gcc-ranlib" % toolchain_arch,
     SIZETOOL="%s-elf-size" % toolchain_arch,
 
     ARFLAGS=["rc"],


Jason2866:
@TheRealMazur  Great find! Just tried and compiled on Apple M1 and the flash size reduced from 1.427.984 bytes to 1.276.720 bytes. **147k less flash** used.

TheRealMazur:
@Jason2866 just to be sure - you tried applying my changes and enabling lto, right? This PR should not have any effects without enabling lto AFAIK 😅

Jason2866:
Yes, since I forked this github, and espressif Arduino32 for our needs already it was a easy change to use the option lot for compile and linking

TheRealMazur:
Just keep in mind that this can mess up backtrace symbols. I needed to add `-ggdb` linker flag to have the full debug info, which surprisingly increased the flash usage (the end result was still better than without lto, but I was sure any debug symbols will be stripped before flashing).

Jason2866:
For me using `-flto` for C3 does not work. Ends in bootloop.
S2 and ESP32 does work without any problems.

TheRealMazur:
It's possible that some specific linker flags or scripts would need to be modified to work with C3. That shouldn't block this PR, as it merely allows for opting in to use lto.

Unfortunately I'm not able to help you with with debugging C3, as I don't own one.

Jason2866:
yes, this PR should be merged!

krzmaz:
@valeros @ivankravets Sorry for the direct pings but I'm not sure how to get some progress on this PR. Could you help move it forward?

Jason2866:
We use in our fork for Tasmota. `lto` works great, excluding the c3 (known gcc compiler bug, solved in actual gcc version).

Jason2866:
Bump

ivankravets:
This is an official dev-platform and we try to keep it in accordance with official Espressif build workflows. Nevertheless, you can customize the toolchain for your project by leveraging [PRE scripting](https://docs.platformio.org/en/latest/scripting/launch_types.html):

**platformio.ini**

```ini
[env]
extra_scripts = pre:custom_toolchain.py
```

**custom_toolchain.py**

```python

Import("env")

env.Replace(
    AR="xtensa-esp32-elf-gcc-ar",
    RANLIB="xtensa-esp32-elf-ranlib"
)

```

Could you try the snippet above?

Jason2866:
Good reason. @me-no-dev can we switch to this versions?

me-no-dev:
@Jason2866 feel free to PR to the 5.1-libs branch. We can have it as part of 3.0.0

softhack007:
+1 👍 for merging this into development.

zackees:
This issue also affects AVR platforms. We're seeing it in FastLED where large static arrays from third-party libraries (stb_vorbis, TJpg_Decoder, etc.) are being linked into the final binary even though they're completely unreachable.

On ATtiny1616 (2KB RAM), the .bss section overflows by ~1.5KB because unused lookup tables like `crc_table[256]` and `inverse_db_table[256]` aren't being garbage collected - even with `-flto -ffunction-sections -fdata-sections -Wl,--gc-sections`.

The root cause is that standard `ar`/`ranlib` don't understand LTO bytecode (GIMPLE), so the static library archive doesn't participate properly in link-time optimization. When we completely exclude the third-party libraries, .bss drops from 3620 bytes to just 94 bytes.

Would be great to see `gcc-ar`/`gcc-ranlib` support extended to the AVR platform as well (`platform-atmelavr` / `platform-atmelmegaavr`).

zackees:
**Update: Alternative workaround discovered**

While investigating this issue for FastLED, we found an alternative approach that avoids the `ar` vs `gcc-ar` problem entirely:

**Pass object files directly to the linker instead of archiving them.**

When using LTO with `-fno-fat-lto-objects`, the compiled objects contain only GIMPLE bytecode (no machine code). Standard `ar` can't build a proper symbol index for these objects, so the linker can't eliminate unused code/data from the archive.

The workaround implemented in [fbuild](https://github.com/zackees/fbuild) (an alternative Arduino build system):

```python
# From fbuild/src/fbuild/build/linker.py:
# Note: Core objects are passed directly instead of being archived because
# LTO with -fno-fat-lto-objects produces bytecode-only objects that don't
# work well in archives (the archive won't have a proper symbol index).

link_result = linker.link_legacy(
    sketch_objects,
    core_objects,      # Passed directly, not archived
    output_elf,
    [],                # No library archives
    None,
    lib_objects        # Library objects passed separately for LTO
)
```

This allows LTO to work properly because the linker sees the GIMPLE bytecode directly rather than through an archive.

For PlatformIO, either:
1. Switch to `gcc-ar` (which understands GIMPLE and builds proper symbol indices), or
2. Pass library object files directly to the linker instead of archiving them

Both approaches would fix the issue where unused static data in libraries cannot be eliminated by the linker.

Jason2866:
pioarduino does use the LTO capable gcc binaries. ArduinoIDE cant use since the build engine cant handle.

zackees:
An LTO enabled build through hacks is quite startling. The attiny85 on fastled shaves off **2.5k** of ram usage from final program linkage with libfastled.a (added due to a new feature) bringing it down to **94** bytes.

zackees:
@Jason2866 i'm not seeing LTO on on our esp32dev which may or may not be using pioarduino, is this expected?

I'm seeing a third party jpeg library make it into the build for a Blink.ino sketch despite no references to. I've confirm that it get's stripped out with LTO enabled.

Edit: Our esp32dev is an idf 4 build. I see the patch went in for idf 5

Jason2866:
@zackees using LTO with IDF 4.4 toolchains is not a good idea. The riscv generates faulty code

zackees:
@Jason2866 if the compiler toolchain was updated for 4.4 would this be solved?

Jason2866:
@zackees afaik the bug is solved around gcc >12 as the base for the build toolchains. Since IDF 4.4 uses gcc 8.x there is no way to get a working LTO for riscv here. IDF 4.4 is a dead horse. Out of support from espressif.

Issue #619, opened 2021-09-09
Support for the latest Arduino v2.0


valeros:
The latest Arduino core adds support for ESP32S2, ESP32C3 SoCs and a great amount of new boards. Related issues:
- https://github.com/platformio/platform-espressif32/issues/591
- https://github.com/platformio/platform-espressif32/issues/557
- https://github.com/platformio/platform-espressif32/issues/564
- https://github.com/platformio/platform-espressif32/issues/570
- https://github.com/platformio/platform-espressif32/issues/546
- https://github.com/platformio/platform-espressif32/issues/521
- https://github.com/platformio/platform-espressif32/issues/609
- https://github.com/espressif/arduino-esp32/issues/5255

MichaelDu9226:
Hi all
Are there any plans for this feature? When will it be supported?

maxgerhardt:
You might also want to take a look at https://github.com/espressif/arduino-esp32/issues/5505 in regards to this.

forkineye:
> I've been trying to understand how I can get Arduino v2.0.0 to work in PlatformIO because there are some important bugfixes in esp-idf v4.3 which I need.. I have tried building with Arduino as an IDF component but that only supports up to IDF v4.0 and is buggy. I can't think of a way to get these updates without a lot of rework on my project by converting entirely to IDF.
> 
> Maybe it is my lack of experience with the platform, but I am confused about what prevents us using Arduino v2.0 in PlatformIO, can you explain it for me?

Here's an example that may help you out.  The relevant tidbits are in [esp32git] which just extends the [esp32] entry I have an overrides the platform until support is added.

```ini
;~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~;
; ESP32 defaults for 4MB flash                                     ;
; https://docs.platformio.org/en/latest/platforms/espressif32.html ;
;~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~;
[esp32]
platform = espressif32
board_build.filesystem = littlefs
board_build.partitions = min_spiffs.csv
monitor_filters = esp32_exception_decoder
build_flags =
    ${env.build_flags}
lib_deps =
    ${env.lib_deps}
    esphome/AsyncTCP-esphome @ 1.2.2

[...]

PR #714, opened 2022-01-28
Use correct object copy for TxtToBin

maxgerhardt:
See https://github.com/platformio/platform-espressif32/issues/619#issuecomment-1024459499


valeros:
Many thanks!

Issue #644, opened 2021-10-11
Add support for DFRobot FireBeetle 2 (ESP-WROOM-32E)


alex-code:
Please add support for FireBeetle 2, https://www.dfrobot.com/product-2195.html
The updated board uses an ESP-WROOM-32E

matty0005:
I second this - any updates on this?

mereshow:
I second this, also.

k2m5t2:
Any updates?

valeros:
Unfortunately, this board is not supported by the official Arduino core for ESP32. It might be a good idea to add it here https://github.com/espressif/arduino-esp32 in the first place.

jyhw123:
Not sure if this helps anyone but after changing my platformio.ini file to 
```
[env:firebeetle32]
platform = espressif32@2.1.0
board = firebeetle32
framework = arduino
monitor_speed = 115200
board_build.f_flash = 80000000L
board_build.flash_mode = dio
```
allowed me to finally flash some code and read serial output from my Firebeetle ESP32-E (SKU:DFR0654)

imabot2:
@jyhw123 I try your config, I can program the ESP32, pin out seems to be wrong :

```
src/main.cpp:4:24: error: 'A2' was not declared in this scope
 const int BATTERYPIN = A2;
```

lmarzen:
I was able to upload to my Firebeetle 2 ESP32-E earlier last year (June 2022) using `platform = espressif32`.  It seems like v5.1.0 caused issues for that solution. Use `platform = espressif32@5.0.0` instead.

Dozentz:
Ok

lmarzen:
In order to upload code to the Firebeetle 2 ESP32-E using the latest version of espressif32, you must override 3 of the default settings for the firebeetle32 board configuration. See my platformio.ini file below.
```
[env:firebeetle32]
platform = espressif32
board = firebeetle32
framework = arduino

; override default setting to enable compatibility with Firebeetle 2 ESP32-E
board_upload.flash_size = 4MB
board_upload.maximum_ram_size = 327680
board_upload.maximum_size = 4194304
```

PR #745, opened 2022-03-23
Add firebeetle32e board

Wumpf:
And correct links as well as flash size for firebeetle32

Product page has vague comparison between the two boards.
https://www.dfrobot.com/product-2231.html
But I opted to put in the dfrobot wiki pages nontheless as they are generally more useful.

Was not quite sure about the `maximum_ram_size` to put in - pretty much all boards advertise 520k but all are listed here with 327680, so I reckon there's some reason for that :)

### DIFF
--- a/boards/firebeetle32e.json
+++ b/boards/firebeetle32e.json
@@ -9,7 +9,7 @@
     "f_flash": "40000000L",
     "flash_mode": "dio",
     "mcu": "esp32",
-    "variant": "firebeetle32"
+    "variant": "firebeetle32e"
   },
   "connectivity": [
     "wifi",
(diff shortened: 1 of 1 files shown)


Wumpf:
just saw this fixes #644
Although I didn't do anything about it being esp-wroom-32e rather than esp-wroom-32. Don't think it's needed but I'm new to this :)

Jason2866:
All entries with ram sizes different to `4194304` for esp32 are just wrong

Wumpf:
@Jason2866 you refer to `maximum_ram_size` or to `maximum_size` (max program size..?). Maximum size is speced to 4194304 everywhere and reverted that now here as well. But why is that? E.g. rom size here is speced with 448kb https://www.espressif.com/sites/default/files/documentation/esp32-wroom-32e_esp32-wroom-32ue_datasheet_en.pdf
Also, as mention can someone explain why ram sizes are always set to 327680, again independently of the 520k mentioned in spec sheets?

Jason2866:
Since the 520k are not useable with IDF / Arduino. Max useable RAM is the same for ALL ESP32.

valeros:
Hi @Wumpf, thanks for the PR. Am I missing something, or this board is not supported by the official Arduino core for ESP32 ? There is simply no such variant `firebeetle32e` here https://github.com/espressif/arduino-esp32/tree/master/variants

Wumpf:
oh, you're right! Also on the tutorial on the DFRobot [wiki](https://wiki.dfrobot.com/FireBeetle_Board_ESP32_E_SKU_DFR0654#target_6) they use just firebeetle as board type in the arduino ide

PR #752, opened 2022-03-31
BPI runs at 240MHz

floitsch:
It looks like the BPI bit runs with a wroom-32 module which generally runs at 240MHz.
Also, the documentation mentions 240MHz several times: https://wiki.banana-pi.org/BPI-Bit

### DIFF
--- a/boards/bpi-bit.json
+++ b/boards/bpi-bit.json
@@ -5,7 +5,7 @@
     },
     "core": "esp32",
     "extra_flags": "-DARDUINO_BPI_BIT",
-    "f_cpu": "160000000L",
+    "f_cpu": "240000000L",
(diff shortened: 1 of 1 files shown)


floitsch:
Fwiw, I don't own this board and was not able to test the PR.

valeros:
Hi @floitsch ! Thanks for the PR. Arduino framework specifies 160MHz for this board. Does it make sense to fix the value there in the first place?

floitsch:
I think that would make sense.
I don't own any bpi-bit, so someone should probably verify it first...

Do you have a link to the Arduino framework with the bpi-bit?

valeros:
https://github.com/espressif/arduino-esp32/blob/master/boards.txt#L10135

PR #775, opened 2022-04-26
PSRAM fix not needed

Jason2866:
since this revision is not affected from

@hallard


hallard:
Awesome, thanks for PR'ing for this board :-)

valeros:
Thanks!

Issue #780, opened 2022-04-27
Add support for generic Esp32 WRoom devkit with 16Mb 


Bascy:
Config is the same as in `esp32dev.json` with the following changes:

`
  "flash_size": "16MB",
  "maximum_size": "maximum_size": 16777216,
`

valeros:
Hi @Bascy ! Simply use the default `esp32dev` ID and override the `flash_size` and `partitions` options in your `platformio.ini`:
```ini
platform = espressif32
board = esp32dev
framework = arduino

board_build.partitions = default_16MB.csv
board_upload.flash_size = 16MB
```

PR #781, opened 2022-04-27
Added generic esp32 Devkit 16Mb configuration (issue #780)

Bascy:


valeros:
Hi @Bascy ! There is no need for an additional generic board as you can use the default `esp32dev` ID and override the `flash_size` and `partitions` options in your `platformio.ini`:
```ini
platform = espressif32
board = esp32dev
framework = arduino

board_build.partitions = default_16MB.csv
board_upload.flash_size = 16MB
```

PR #785, opened 2022-04-30
remove fixed address 0x10000 for firmware

Jason2866:
since this prevents to use a factory firmware which should be placed at address 0x10000.
The flash address for the user firmware (ota_0) is read from the [partition.csv](https://github.com/arendst/Tasmota/blob/development/partitions/esp32_partition_app1856k_fs1344k.csv).
A [flash_extra_images](https://github.com/arendst/Tasmota/blob/development/boards/esp32_4M.json) as example for useage.
Idk what this part of the code has done before.


Jason2866:
bump

proddy:
+1

Jason2866:
@valeros I noticed that setting the env variable here does not work as expected https://github.com/platformio/platform-espressif32/blob/ee6f97db592c3697697bcfe76aba66ce0c64c329/builder/main.py#L105
It seems it is encapsulated. Is there a way to set it global? The PR does work nice for setting the address (dynamically) for the firmware when flashing with esptool.py but it fails when using the debugger.

Jason2866:
Updated and debugger now uses the correct offset too. No need to specify `ESP32_APP_OFFSET` anywhere since it is read from the partitions.csv

PR #807, opened 2022-05-17
boards/um_feathers3: Don't flash tinyuf2.bin

prplz:
In Arduino IDE, tinyuf2.bin is only flashed when the tinyuf2 partition option is selected from the menu. Flashing the tinyuf2 binary without the appropriate partitions does not work and clobbers part of the app0 partition. cc @UnexpectedMaker


valeros:
Many thanks, merged!

PR #829, opened 2022-06-11
Update for DFRobot ESP32 boards

doskoi:
* Add new product Beetle ESP32-C3
* Fix the wrong config from Firebeetle ESP32


valeros:
Hi @doskoi , thanks for the PR. I don't see the `beetle32c3` variant present in the official ESP32 Arduino core https://github.com/espressif/arduino-esp32/tree/master/variants. Please add your variant to that repository and once it's available in the stable release it will be added to PlatformIO.

boards/beetle32-c3.json
```diff
@@ -0,0 +1,39 @@
...
+      "arduino":{
+        "ldscript": "esp32c3_out.ld"
+      },
+      "core": "esp32",
+      "f_cpu": "160000000L",
+      "f_flash": "80000000L",
+      "flash_mode": "qio",
+      "extra_flags": [
+        "-DARDUINO_ESP32C3_DEV",
+        "-DARDUINO_USB_CDC_ON_BOOT=1"
```
unlogisch04, line 12: In framework-arduinoespressif32\platform.txt the ARDUINO_USB_MODE is set. Without it i had no "Serial".
```suggestion
        "-DARDUINO_ESP32C3_DEV",
        "-DARDUINO_USB_MODE=1",
        "-DARDUINO_USB_CDC_ON_BOOT=1"
```

unlogisch04:
What was there a reason to remove` "-DARDUINO_USB_CDC_ON_BOOT=1" `? 
Currently i only get when i use Serial.println("abc"); nothing on the output. But i get debug output. `[  9515][E][WiFiSTA.cpp:317] ....`   with the ` "-DARDUINO_USB_CDC_ON_BOOT=1" ` i get the expected output.
The USB Serial Port is directly connected to the ESP Chip.

valeros:
We tend to use the same settings as in the Arduino IDE. By default, the CDC is disabled there https://github.com/espressif/arduino-esp32/blob/master/boards.txt#L6218. If you think it's a mistake, feel free to provide a PR.

PR #830, opened 2022-06-11
Add hwids for OLIMEX ESP32 boards that use CH340 USB/serial converters

egnor:
I've verified in olimex's site that these boards all use the CH340 USB/serial converter, which has USB VID=1A86 PID=7523. While not unique (lots of devices use this converter), listing it is still helpful for distinguishing from other serial devices.

Note that the olimex ESP32-PRO does _not_ use this converter, instead it has a PIC32 used for USB, so it's not included in this change. (I don't have one, and I'm not sure what VID/PID it would end up with.)


valeros:
Many thanks, merged.

PR #687, opened 2021-12-14
esp32_exception_decoder: fix decoding when space is missing

mhaberler:
Host: MacOS Big Sur 11.6
Target: M5Stack core2
platformio.ini:    
    monitor_filters = direct,   esp32_exception_decoder

The backtrace message lacks a space after the top-of-stack address,
consequently the backtrace regex does not trigger and backtrace not shown.

> Backtrace:0x40083161:0x3ffb24700x40089315:0x3ffb2490 0x4008de89:0x3ffb24b0
> ------ missing space --------------^ 

by replacing the  '+' (1..n )length of hex addresses by fixed repeat count of 8, 
these backtraces can be made to decode correctly.

While the error in the message seems specific to the M5Stack Core2, the patch does no harm if the format is correct.

as per 495c689133146ef8d4520d8ff3b10a720508a0c8  - backtrace not decoded:
```
> Executing task: platformio device monitor <

source /Users/mah/.platformio/penv/bin/activate
--- Available filters and text transformations: colorize, debug, default, direct, esp32_exception_decoder, hexlify, log2file, mah_exception_decoder, nocontrol, printable, send_on_enter, time
--- More details at https://bit.ly/pio-monitor-filters
Esp32ExceptionDecoder: firmware at /Users/mah/Documents/PlatformIO/Projects-private/m5core2-lvgl-demo/.pio/build/m5stack-core2/firmware.elf does not exist, rebuild the project?
--- Miniterm on /dev/cu.SLAB_USBtoUART  115200,8,N,1 ---
--- Quit: Ctrl+C | Menu: Ctrl+T | Help: Ctrl+T followed by Ctrl+H ---
Hello Arduino! V8.1.1
I am LVGL_Arduino
Setup done

assert failed: xQueueGenericSend queue.c:820 (pxQueue)


Backtrace:0x40083161:0x3ffb24700x40089315:0x3ffb2490 0x4008de89:0x3ffb24b0 0x40089ade:0x3ffb25e0 0x4011367a:0x3ffb2620 0x401137b5:0x3ffb2640 0x400d5e27:0x3ffb2660 0x400d5ed2:0x3ffb2680 0x400d6008:0x3ffb26d0 0x400d147c:0x3ffb2720 0x400e98c5:0x3ffb2760 0x400d7fc9:0x3ffb2780 0x400eac62:0x3ffb27c0 0x400ead0a:0x3ffb27e0 0x400d169f:0x3ffb2800 0x40104ae5:0x3ffb2820 




[...]

PR #831, opened 2022-06-11
Better fix for esp32_exception_decoder missing space

MitchBradley:
Recent versions of ESP-IDF have a bug in which the Backtrace: line is
malformed.

Old/Good: Backtrace: 0xN:0xM 0xN:0xM 0xN:0xM ...
New/Bad:  Backtrace:0xN:0xM0xN:0xM 0xN:0xM ...

I issued https://github.com/espressif/esp-idf/pull/9138 against
ESP-IDF to fix the underlying problem, but it is unclear when or if
that PR will be accepted, especially since the monitor program in
ESP-IDF is immune to the problem.

By using a decoding technique similar to the one in ESP-IDF's monitor,
either backtrace format can be accepted.

Supersedes #687 , which does not quite work, in that it misses one of
the entries.

### DIFF
--- a/monitor/filter_exception_decoder.py
+++ b/monitor/filter_exception_decoder.py
@@ -32,11 +32,11 @@ IS_WINDOWS = sys.platform.startswith("win")
 class Esp32ExceptionDecoder(DeviceMonitorFilterBase):
     NAME = "esp32_exception_decoder"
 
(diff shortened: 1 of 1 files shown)


MitchBradley:
Please note that the structure of this patch, with the definitions of BACKTRACE_PATTERN and BACKTRACE_ADDRESS_PATTERN, is intentionally similar to tools/ci/python_packages/ttfw_idf/IDFDUT.py in the ESP-IDF repo.

ivankravets:
Thanks for the PR!

PR #836, opened 2022-06-17
Board definition adafruit_feather_esp32s3.json: change "flash_size" property to 8MB

idea--list:
This particular board has 8MB flash with no PSRAM. If the property is set to 4MB as in the current release version, platformio will throw an error while trying to flash any firmware to the board.

Setting flash_size to 8MB fixes the issue.

### DIFF
--- a/boards/adafruit_feather_esp32s3.json
+++ b/boards/adafruit_feather_esp32s3.json
@@ -50,7 +50,7 @@
         ]
       ]
     },
-    "flash_size": "4MB",
+    "flash_size": "8MB",
     "maximum_ram_size": 327680,
     "maximum_size": 4194304,
     "use_1200bps_touch": true,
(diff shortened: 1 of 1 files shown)


boards/adafruit_feather_esp32s3.json
```diff
@@ -50,7 +50,7 @@
         ]
       ]
     },
-    "flash_size": "4MB",
+    "flash_size": "8MB",
     "maximum_ram_size": 327680,
     "maximum_size": 4194304,
```
idea--list, line 55: also "maximum_size" shall be changed to 8388608

Jason2866:
There are two boards variants availabe sold with the same name.
https://www.adafruit.com/products/5323
https://www.adafruit.com/products/5477
So it needs a more precise definition in the name

Jason2866:
Adafruit opened a PR to add more variants https://github.com/espressif/arduino-esp32/pull/6881

valeros:
Thanks for the PR, merged.

PR #851, opened 2022-07-12
add Adafruit Feather ESP32-S3 TFT board

egnor:
Product page: https://www.adafruit.com/product/5483
Arduino boards.txt entry: https://github.com/espressif/arduino-esp32/blob/c93bf11f1e3c3ca8088b837525dcdd78c8579a7a/boards.txt#L7984
Arduino variant definition: https://github.com/espressif/arduino-esp32/tree/master/variants/adafruit_feather_esp32s3_tft

Note this is a refresh of the Adafruit Feather ESP32-**S2** TFT board.

### DIFF
--- a/boards/adafruit_feather_esp32s3_tft.json
+++ b/boards/adafruit_feather_esp32s3_tft.json
@@ -1,25 +1,32 @@
 {
   "build": {
     "arduino": {
-      "ldscript": "esp32s3_out.ld"
+      "ldscript": "esp32s3_out.ld",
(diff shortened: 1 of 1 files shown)


egnor:
It's worth noting, this requires the current head version of arduino-esp32 (for the arduino platform anyway).

egnor:
(By the way, I just added a second commit to add another VID:PID that is what the board shows up as when it's in the bootloader.)

PR #869, opened 2022-08-03
wrong flash mode resulting in issues

Jason2866:
issue https://github.com/espressif/arduino-esp32/issues/6572#issuecomment-1203630400
similar the fix for the S3 esp32-s3-devkitc-1.json

### DIFF
--- a/boards/pico32.json
+++ b/boards/pico32.json
@@ -7,7 +7,7 @@
     "extra_flags": "-DARDUINO_ESP32_PICO",
     "f_cpu": "240000000L",
     "f_flash": "40000000L",
-    "flash_mode": "dio",
+    "flash_mode": "qio",
     "mcu": "esp32",
     "variant": "pico32"
   },


Jason2866:
@valeros Probably a change for all newer boards to flash mode `qio` will avoid more issues here. Only old ESP32 board designs may not support mode `qio`.

valeros:
Hi @Jason2866 ! Thanks for the PR. I'm in favor of keeping the flash mode settings in sync with the Arduino core for ESP32. For example, they use the `dio` mode for the `ESP32 Pico Kit` board https://github.com/espressif/arduino-esp32/blob/master/boards.txt#L1014:L1015. Is it a mistake?

Jason2866:
Well, than the question is why `dio` works with Arduino IDE and it fails with PlatformIO?

Jason2866:
Closing [this PR ](https://github.com/espressif/esp32-arduino-lib-builder/pull/80) will fix  the issues with flash modes. All modes are possible.

PR #880, opened 2022-08-14
fix bootloader flash mode

Jason2866:
since it is not possible to flash in boot mode `qio` / `qout`


TD-er:
Maybe also add PIO option to set other flags like not adding SHA digest?

Jason2866:
@TD-er The command not to add sha diggest is introduced in esptool v4.x.
Adding without updating to esptool v.4x would result in a error.

valeros:
Thanks, merged!

PR #887, opened 2022-08-20
Denky D4 board has better specs

Jason2866:
@hallard fyi
@valeros @hallard is the creator of this board ;-)


hallard:
@Jason2866 thanks 
and latest version has all components on same side, USB-C and QWIIC connector :-)
image

valeros:
Many thanks!

PR #888, opened 2022-08-22
Add USB_CDC flag to Beetle ESP32-C3

fredlarochelle:
Adding "-DARDUINO_USB_CDC_ON_BOOT=1" to dfrobot_beetle_esp32c3.json, otherwise, the serial monitor from USB doesn't work.


valeros:
Thanks, merged!

PR #890, opened 2022-08-22
set specific variant for lolin_c3_mini

xsrf:
I noticed that there is a specific variant for `lolin_c3_mini` but it is not used e.g. when compiling for the arduino framework. This caused e.g. `LED_BUILTIN` to resolve to the wrong pin number. This should fix it, I guess.


valeros:
Thanks!

PR #901, opened 2022-09-03
Latest Arduino master does fail with CI

Jason2866:
since example sdkconfig settings needs to be changend


valeros:
Many thanks, merged!

Issue #837, opened 2022-06-18
esp32-s3-devkitc-1 issue


fabiencomte:
Hello,
Basic hello world doesn't work.

#include "Arduino.h"

void setup() {
  delay(1000);
  Serial.begin(115200);
  Serial.print("Hello world\n");
}

void loop() {
    delay(1000);
    Serial.print("Hello world\n");
}

This is the monitor output.

Rebooting...
␀�ESP-ROM:esp32s3-20210327
Build:Mar 27 2021
rst:0xc (RTC_SW_CPU_RST),boot:0x18 (SPI_FAST_FLASH_BOOT)
Saved PC:0x40376c18
SPIWP:0xee
Octal Flash Mode Enabled
For OPI Flash, Use Default Flash Boot Mode
mode:SLOW_RD, clock div:1
load:0x3fcd0108,len:0x39c
load:0x403b6000,len:0x9a4
load:0x403ba000,len:0x2868
entry 0x403b61c0

assert failed: do_core_init startup.c:298 (flash_ret == ESP_OK)

Backtrace:0x403769f6:0x3fceb1800x40379b35:0x3fceb1a0 0x4037f095:0x3fceb1c0 0x42007fc2:0x3fceb2f0 0x40376511:0x3fceb320 0x403bae67:0x3fceb350 0x403bb0ee:0x3fceb380 0x403b6215:0x3fceb4b0 0x40045c01:0x3fceb570  |<-CORRUPTED

ELF file SHA256: 0000000000000000

Rebooting...
␀�ESP-ROM:esp32s3-20210327
Build:Mar 27 2021
rst:0xc (RTC_SW_CPU_RST),boot:0x18 (SPI_FAST_FLASH_BOOT)
Saved PC:0x40376c18
SPIWP:0xee
Octal Flash Mode Enabled
For OPI Flash, Use Default Flash Boot Mode
mode:SLOW_RD, clock div:1
load:0x3fcd0108,len:0x39c
load:0x403b6000,len:0x9a4
load:0x403ba000,len:0x2868
entry 0x403b61c0

assert failed: do_core_init startup.c:298 (flash_ret == ESP_OK)

[...]

PR #904, opened 2022-09-04
Fix flash mode for `opi`

Jason2866:
opi needs to be flashed in mode `dout` to work. Changed flash mode for `qout` since it can be flashed in `dio`.
This is the way espressif now does in IDF
Related https://github.com/platformio/platform-espressif32/issues/837#issuecomment-1236378086

### DIFF
--- a/builder/main.py
+++ b/builder/main.py
@@ -51,10 +51,10 @@ def _get_board_f_flash(env):
 def _get_board_flash_mode(env):
     memory_type = env.BoardConfig().get("build.arduino.memory_type", "qio_qspi")
     mode = env.subst("$BOARD_FLASH_MODE")
-    if mode == "qio" or mode == "qout":
-        return "dio"
     if memory_type == "opi_opi" or memory_type == "opi_qspi":
         return "dout"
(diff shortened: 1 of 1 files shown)


valeros:
Hi @Jason2866 ! Thanks for the PR. I'm curious how the priority of these checks should be organized. Shouldn't the `if memory_type == "opi_opi" or memory_type == "opi_qspi"` be evaluated first?

> Changed flash mode for qout since it can be flashed in dio. This is the way espressif now does in IDF

Could you please point me to the code where it's handled in IDF?

Jason2866:
☺valeros [Here](https://github.com/espressif/esp-idf/blob/master/components/esptool_py/Kconfig.projbuild) are the flash modes set

Jason2866:
> Shouldn't the if memory_type == "opi_opi" or memory_type == "opi_qspi" be evaluated first?

No since this needs to override all other settings before

Jason2866:
@valeros changed

valeros:
Thanks, merged!

Jason2866:
Hi @valeros The PR #880 fixes another issue.

PR #907, opened 2022-09-05
Added bootloader override option

tst-fractal:
Added bootloader override option, the IDE now checks if the "bootloader_components" directory is present and tries to compile using the new bootloader provided.  
Added example for usage.   
Please note:  
in examples/espidf-custom-bootloader/platfromio.ini  

platform = https://github.com/tst-fractal/platform-espressif32.git#develop **must be changed to**   
platform = espressif32  when merging


valeros:
Hi @tst-fractal ! Many thanks for a great PR. Upon first sight, I'm a bit worried about hardcoded paths like:
```python
bootloader_subproject_dir = os.path.join(
  PROJECT_DIR, "bootloader_components", "subproject"
)
```
It works fine when you just copy the same bootloader from the framework package, but what if a user needs an arbitrary structure? For instance, there is an official project example with [a custom bootloader](https://github.com/espressif/esp-idf/tree/master/examples/custom_bootloader/bootloader_override) and I believe your current implementation won't be able to compile it successfully. Any thoughts?

PR #913, opened 2022-09-14
Use correct bootloader `opi` when `"boot": "opi"` is not set

Jason2866:
for ESP32-S3 devices with flash type `opi`.  This prevents errors in compile since platformio searches for the bootloader `bootloader_dout_80m.bin` which does not exists and would be wrong.
With the change the correct bootloader `bootloader_opi_80m.bin` is used.

@chegewara thx for pointing at and this PR [fixes](https://github.com/platformio/platform-espressif32/issues/837#issuecomment-1246382561) this issue.

@valeros enhances #904 since it is more user friendly. It is hard to know to set all needed options correctly.
This is now done automatically

### DIFF
--- a/builder/main.py
+++ b/builder/main.py
@@ -59,9 +59,9 @@ def _get_board_flash_mode(env):
 
 
 def _get_board_boot_mode(env):
-    memory_type = env.BoardConfig().get("build.arduino.memory_type")
+    memory_type = env.BoardConfig().get("build.arduino.memory_type", "")
     build_boot = env.BoardConfig().get("build.boot", "$BOARD_FLASH_MODE")
-    if memory_type in ("opi_opi", "opi_qspi"):
+    if ["arduino"] == env.get("PIOFRAMEWORK") and memory_type in ("opi_opi", "opi_qspi"):
         build_boot = "opi"
     return build_boot
 
--- a/builder/main.py
+++ b/builder/main.py
@@ -49,7 +49,7 @@ def _get_board_f_flash(env):
 
 
 def _get_board_flash_mode(env):
-    memory_type = env.BoardConfig().get("build.arduino.memory_type")
+    memory_type = env.BoardConfig().get("build.arduino.memory_type", "qio_qspi")
     mode = env.subst("$BOARD_FLASH_MODE")
     if memory_type in ("opi_opi", "opi_qspi"):
         return "dout"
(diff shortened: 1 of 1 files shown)


valeros:
Hi @Jason2866 ! Thanks for the PR!  You're trying to evaluate an Arduino-specific option in the generic `_get_board_boot_mode` function that may be used with other frameworks, so I'm wondering how this could possibly affect IDF-based projects. Should these changes be only applied to Arduino-based projects?

Jason2866:
Hi @valeros i dont see how this could affect other frameworks since `memory_type = env.BoardConfig().get("build.arduino.memory_type", "qio_qspi")` is only set in Arduino env and so the next lines will do no changes in other frameworks

valeros:
> i dont see how this could affect other frameworks since memory_type = env.BoardConfig().get("build.arduino.memory_type", "qio_qspi") is only set in Arduino env and so the next lines will do no changes in other frameworks

I'm afraid that's not accurate assumption. The `build.arduino.memory_type` option is available regardless of the used framework. Besides, you specified a default value `qio_qspi` which will be used when this setting is not set.

Jason2866:
Talking about not compatible stuff for frameworks, this https://github.com/platformio/platform-espressif32/blob/develop/builder/main.py#L224 is bad for IDF and Arduino.
Better read from partitions.csv I use #785 in my fork and it does work nice

Jason2866:
> I'm afraid that's not accurate assumption. The build.arduino.memory_type option is available regardless of the used framework. Besides, you specified a default value qio_qspi which will be used when this setting is not set.

Everytime open for a better solution. The default can be removed. Any way to set correct bootloader `opi` without needeing a entry will help.

valeros:
> Everytime open for a better solution. The default can be removed. Any way to set correct bootloader opi without needeing a entry will help.

I guess removing default won't be enough, as there might be boards that have this option in their manifests by default, but used in IDF-based projects. If these changes are not applicable to IDF, can we just check whether the current project is Arduino only, e.g.:

```py
memory_type = env.BoardConfig().get("build.arduino.memory_type", "")
if ["arduino"] == env.get("PIOFRAMEWORK") and memory_type  in ("opi_opi", "opi_qspi"):
  return "dout"
...
```

BTW, should these settings also be used for mixed IDF/Arduino projects?

Jason2866:
Yes, the setting should be use in mixed in Arduino/IDF projects.

PR #914, opened 2022-09-14
fix "memory_type" for esp32s3box

### LINKED ISSUE #912: esp32s3box  sdkconfig.h  no such file
### Actual Results
image

### If problems with PlatformIO Build System:

**The content of `platformio.ini`:**
```ini
[env:esp32s3box]
platform = espressif32
board = esp32s3box
framework = arduino
```

**Source file to reproduce issue:**
```cpp
#include <Arduino.h>

void setup() {
  // put your setup code here, to run once:
}

void loop() {
  // put your main code here, to run repeatedly:
}
```

Jason2866:
since the naming changed. Fixes #912
Remove not needed entrys

### DIFF
--- a/boards/esp32s3box.json
+++ b/boards/esp32s3box.json
@@ -2,12 +2,10 @@
   "build": {
     "arduino":{
       "ldscript": "esp32s3_out.ld",
-      "memory_type": "qspi_opi"
+      "memory_type": "qio_opi"
     },
     "core": "esp32",
     "extra_flags": [
-      "-DARDUINO_RUNNING_CORE=1",
-      "-DARDUINO_EVENT_RUNNING_CORE=1",
       "-DBOARD_HAS_PSRAM",
       "-DARDUINO_ESP32_S3_BOX",
       "-DARDUINO_USB_MODE=1",


valeros:
Hi @Jason2866 ! Thanks for the PR. Are you sure the ` -DARDUINO_RUNNING_CORE=1`, `-DARDUINO_EVENT_RUNNING_CORE=1` flags are not needed? It seems they are also added in the Arduino IDE.

Jason2866:
@valeros The are set [here](https://github.com/espressif/arduino-esp32/blob/master/Kconfig.projbuild#L79-L99) and this is resulting in [sdkconfig.h](https://github.com/espressif/arduino-esp32/blob/master/tools/sdk/esp32/qio_qspi/include/sdkconfig.h) as one example. This file is always included when building a Arduino sketch. It is obsolete since this change [commit](https://github.com/espressif/arduino-esp32/commit/09c0a39d2a97e12a6cd01c18e3a7bce8e62bd77f) on 2022/05/13

valeros:
Thanks for the clarification. Still, if it's outdated why they add these flags even in the latest Arduino 2.0.5? Shouldn't this be fixed in the Arduino repository in the first place?

Jason2866:
If you take a look in the boards.txt you will find many outdated and wrong entrys. Since most of the boards are provided via PRs and doing copy&paste from existing boards is common... The settings where needed in the past, so it makes its way in actual new ones. It would be a horrible job to review all existing entrys. AND if a board is used with a old version it will fail. THIS is not the case here, since S3 boards are supported after May.
Btw. Some board manifest in Platformio are not not correct at all.

valeros:
Many thanks!

Jason2866:
@valeros Talking about useless entrys this "maximum_size": 4194304, is a junk entry in Platformio. Since you do it way more clever.  Platformio is reading the partition size and overrides the value here. It could be deleted when the reading of the value in pio code is removed.

Issue #915, opened 2022-09-16
TTGO T-Display-S3 support sanity check


millardjn:
Hello I have been learning to use platformio with a TTGO TDisplayS3 https://github.com/Xinyuan-LilyGO/T-Display-S3

I have managed to get the board working without any checksum errors or issues with my projects so far with the following definition:
```json
{
  "build": {
    "arduino":{
      "ldscript": "esp32s3_out.ld",
      "partitions": "default_16MB.csv",
	  "memory_type": "qio_opi"
    },
    "core": "esp32",
    "extra_flags": [
      "-DARDUINO_ESP32S3_DEV",
      "-DARDUINO_USB_CDC_ON_BOOT=1",
      "-DARDUINO_RUNNING_CORE=1",
      "-DARDUINO_EVENT_RUNNING_CORE=1",
	  "-DBOARD_HAS_PSRAM"
    ],
    "f_cpu": "240000000L",
    "f_flash": "80000000L",
    "flash_mode": "qio",
    "hwids": [
      [
        "0x303A",
        "0x1001"
      ]
    ],
    "mcu": "esp32s3",
    "variant": "esp32s3"
  },
  "connectivity": [
    "wifi"
  ],
  "debug": {
    "default_tool": "esp-builtin",
    "onboard_tools": [
      "esp-builtin"
    ],
    "openocd_target": "esp32s3.cfg"
  },
  "frameworks": [
    "arduino",
    "espidf"
  ],
  "name": "TDisplay Espressif ESP32-S3 (16 MB QD, 8MB PSRAM)",
  "upload": {
    "flash_size": "16MB",
    "maximum_ram_size": 8716288,
    "maximum_size": 16777216,
    "use_1200bps_touch": true,
    "wait_for_upload_port": true,
    "require_upload_port": true,
    "speed": 460800
  },
  "url": "https://github.com/Xinyuan-LilyGO/T-Display-S3",
  "vendor": "Espressif"
}
```
I'm not 100% confident in the above as I have cobbled together this config by copying from similar boards. If anyone can give me some steps to check that the board works correctly with the above I'll be more than happy to test it out so this can be included in the list of supported boards.

Jason2866:
This is not needed:
```
-DARDUINO_RUNNING_CORE=1
-DARDUINO_EVENT_RUNNING_CORE=1
use_1200bps_touch": true,
```
and change entry to
```
maximum_ram_size": 327680,
```
and you can add in the according sections
```
[...]

PR #921, opened 2022-09-24
Add Espressif ESP32-S3-DevKitC-1-N16R8V board file

handledexception:
Tested with PlatformIO 2.5.4 in VSCode with a TTGO T-Display-S3 16MB version.

Should resolve https://github.com/platformio/platform-espressif32/issues/915

### DIFF
--- a/boards/esp32-s3-devkitc-1-n16r8v.json
+++ b/boards/esp32-s3-devkitc-1-n16r8v.json
@@ -1,0 +1,53 @@
+{
+    "build": {
+      "arduino":{
+        "ldscript": "esp32s3_out.ld",
+        "partitions": "default_16MB.csv",
+        "memory_type": "qio_opi"
+      },
+      "core": "esp32",
+      "extra_flags": [
+        "-DARDUINO_ESP32S3_DEV",
+        "-DBOARD_HAS_PSRAM",
+        "-DARDUINO_USB_MODE=1",
+        "-DARDUINO_USB_CDC_ON_BOOT=1"
+      ],
+      "f_cpu": "240000000L",
+      "f_flash": "80000000L",
+      "flash_mode": "qio",
+      "psram_type": "opi",
+      "hwids": [
+        [
+          "0x303A",
+          "0x1001"
+        ]
+      ],
+      "mcu": "esp32s3",
+      "variant": "esp32s3"
+    },
+    "connectivity": [
+      "wifi",
+      "bluetooth"
+    ],
+    "debug": {
+      "default_tool": "esp-builtin",
+      "onboard_tools": [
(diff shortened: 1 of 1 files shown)


handledexception:
Any chance this will get merged?

valeros:
Hi @handledexception, thanks for the PR. IMO, there is no reason to create yet another manifest with minimal changes for a board that's already available. It can be easily configured directly from the project configuration file.

handledexception:
> Hi @handledexception, thanks for the PR. IMO, there is no reason to create yet another manifest with minimal changes for a board that's already available. It can be easily configured directly from the project configuration file.

OK, thank you. I didn't realize that was the intended workflow.

FreeFrags:
@handledexception i just purchased a TTGO T-Display-S3 too and because i couldnt find it in the boards list i created #975.

Did i understand correctly that this issue was closed because we can use a different board and override some properties? 

would you mind telling us how we could get started with the board? im sorry its my first board and im trying to get familiar with the basics

handledexception:
I have no idea what the intended workflow should be to use PlatformIO with the T-Display-S3, hence why I made the PR to add that board file. If there is a workflow that the PlatformIO maintainers intend I would love to hear it. I've just continued using the board file from this PR in my own work and moved on.

thhkd (commented):
Tested in 6.1.11. Worked as expected.

rgov:
> **@valeros:** It can be easily configured directly from the project configuration file.

Would you mind explaining how?

valeros:
@rgov for example:
```ini
[env:esp32-s3-devkitc-1]
platform = espressif32
framework = arduino
board = esp32-s3-devkitc-1

# Configure options for the N16R8V variant
build_flags= -DBOARD_HAS_PSRAM
board_build.arduino.memory_type = qio_opi 
board_build.partitions = default_16MB.csv
board_upload.flash_size = 16MB
```

hsetlik:
Great! Is there a property for specifying the PSRAM size?

rgov:
> Great! Is there a property for specifying the PSRAM size?

I don't believe you have to. I believe the build option just tells the Arduino initialization code to call [`psramInit()`](https://github.com/espressif/arduino-esp32/blob/ccacb7e3d1dd0e58b309c83e5ebc2302ce97c7b2/cores/esp32/esp32-hal-psram.c#L44) which invokes the ESP-IDF function [`esp_psram_init()`](https://github.com/espressif/esp-idf/blob/5f4249357372f209fdd57288265741aaba21a2b1/components/esp_psram/esp_psram.c#L106). Somewhere in there it sends a `PSRAM_DEVICE_ID` command over SPI and uses the reply to determine the PSRAM size.

PR #950, opened 2022-11-01
Sort ULP Sources

DocWibbleyWobbley:
os.listdir returns files in arbitrary order. This can cause relocation errors if files are listed in the "wrong" order.
This sorting allows the user to influence the order via the filenames.


valeros:
Thanks for the PR, merged.

PR #955, opened 2022-11-04
Fix flashing of adafruit_feather_esp32s3_nopsram using espidf

DamarXCV:
Hi,

using espidf i was not able to flash my adafruit_feather_esp32s3_nopsram, with the help of two guys over at the espressif/esp-idf#10026 repo i found out that the "flash_mode": "qio" is not working.

idk why it's working with the arduinoo-framework anyway, my guess is that it's getting overwritten somewhere else to "dio", but i can't find where.

This is probably also an issue with the other adafruit_feather_esp32s3 boards, but i don't have them to test it.

### DIFF
--- a/boards/adafruit_feather_esp32s2_tft.json
+++ b/boards/adafruit_feather_esp32s2_tft.json
@@ -12,7 +12,7 @@
     ],
     "f_cpu": "240000000L",
     "f_flash": "80000000L",
-    "flash_mode": "dio",
+    "flash_mode": "qio",
     "hwids": [
       [
         "0x239A",
--- a/boards/adafruit_feather_esp32s2_tft.json
+++ b/boards/adafruit_feather_esp32s2_tft.json
@@ -12,7 +12,7 @@
     ],
     "f_cpu": "240000000L",
     "f_flash": "80000000L",
-    "flash_mode": "qio",
+    "flash_mode": "dio",
     "hwids": [
       [
         "0x239A",
--- a/boards/adafruit_feather_esp32s3.json
+++ b/boards/adafruit_feather_esp32s3.json
@@ -14,7 +14,7 @@
     ],
     "f_cpu": "240000000L",
     "f_flash": "80000000L",
-    "flash_mode": "qio",
+    "flash_mode": "dio",
     "hwids": [
       [
         "0x239A",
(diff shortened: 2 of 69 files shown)


Jason2866:
The better fix would be to do the correcting for IDF too
https://github.com/platformio/platform-espressif32/blob/develop/builder/main.py#L69
So a simple remove of the check for Arduino would do the job.
NO board can be flashed in mode `qio`

DamarXCV:
Wouldn't it be cleaner to not set the attribute at all (which is probably a bad idea because it has [no default value](https://docs.platformio.org/en/latest/platforms/espressif32.html#flash-mode) right?) nor set a working value in the config even if it gets corrected in the function you linked?

> NO board can be flashed in mode `qio`

Does that mean that there is currently no supported board which supports four wire spi flashing?
Theoretically it's possible with the right chip and wiring, if i understand the [documentation](https://docs.espressif.com/projects/esptool/en/latest/esp32s3/advanced-topics/spi-flash-modes.html#why-don-t-qio-qout-modes-work-with-my-espressif-chip-module) about it correctly!?

DamarXCV:
Is there any reason one would prefer `dout` over `dio` if both are supported by the hardware?

Jason2866:
The second stage bootloader enables the flash mode `qio` when it was builded in mode `qio`. There is no support for `qio` in this early stage. `dout` could be used too. It is not needed since all eap32x boards do support `dio`

Jason2866:
> Wouldn't it be cleaner to not set the attribute at all (which is probably a bad idea because it has [no default value](https://docs.platformio.org/en/latest/platforms/espressif32.html#flash-mode) right?) nor set a working value in the config even if it gets corrected in the function you linked?

Correct values would be always nice

Jason2866:
Your latest commit is NOT correct. This setting is the flash mode used when the sketch is running.

DamarXCV:
Uhm, okey a missunderstanding on my part i guess. My bad!

Jason2866:
This is imho the best way to avoid havoc
https://github.com/Jason2866/platform-espressif32/commit/f9bcd7d3c8ae6a798382cdcfe1525d6176e059d6

DamarXCV:
> This is imho the best way to avoid havoc

Yea, i guess thats true.

> The change is only needed for the S3 !

Want we to change the flash_mode for all S3 or let it be and close the request?

DamarXCV:
> This is imho the best way to avoid havoc [Jason2866@f9bcd7d](https://github.com/Jason2866/platform-espressif32/commit/f9bcd7d3c8ae6a798382cdcfe1525d6176e059d6)

Will you create a pr with the change?

PR #976, opened 2022-12-11
--dont-append-digest

Jason2866:
safeguard to make sure esptool.py v4.4 can patch magic bytes. Since esptool v4.x does check for existing sha diggest and if existing does no header patching


Jason2866:
Closing since not needed anymore, since bootloaders are generated now from *.elf

PR #993, opened 2023-01-07
Fix ULP RISCV compile with IDF5

Jason2866:
now the enabled RISCV ULP is not detected. So wrong FSM ULP setup is done.

@valeros


Jason2866:
Probably wrong for IDF4.4.3 too, since only checked for ESP32-S2.

valeros:
Many thanks, merged!

PR #995, opened 2023-01-10
Expand OS vars in build.esp-idf.sdkconfig_path

gnilk:
This change allows a board configuration to use OS environment variables within the `sdkconfig_path` when building for esp-idf targets.
Like this:
```
    "esp-idf": {
      "sdkconfig_path" : "$MYPROJECT_HOME/PlatformIO/boards/sdkconfig.puckov3_esp32mini.conf"
    },
```


gnilk:
Comment just to avoid closure..
Seems not much happening right now.
Not sure why the unit-tests fails - it is not related to the push as far as I can see it.

PR #1015, opened 2023-01-27
new board LilyGO T-Display-S3

Eistee82:
https://github.com/Xinyuan-LilyGO/T-Display-S3

Pull request new variant: https://github.com/espressif/arduino-esp32/pull/7763


rolupusoru:
ESP32-S3 has 512K RAM not 320KB.

Randomblock1:
also `"partitions": "default_16MB.csv",` should be added in build/arduino to actually use all 16mb

PR #1039, opened 2023-02-20
Added "line" aligned propagation of stream

BOBAH1248:
The additional pre-buffering of input stream with "line aligned" delivery, helps to avoid breaking of escape sequences like "colorization".
It helps for final console output function to not strip "unknown sequences" of non-printable characters.

Now (with my previous commit) colorized output can be used by
```ini
monitor_filters = esp32_exception_decoder
```
i.e. without extra hacks like `monitor_raw = yes` or `monitor_filters = direct, colorize,...`

### DIFF
--- a/monitor/filter_exception_decoder.py
+++ b/monitor/filter_exception_decoder.py
@@ -113,10 +113,6 @@ See https://docs.platformio.org/page/projectconf/build_configurations.html
             last = idx + 1
 
             strip_line = self.strip_ansi(line)
-            # sys.stderr.write(
-            #     "%s: strip_line='%s'\n" % (self.__class__.__name__, strip_line.encode("ascii", "backslashreplace"))
-            # )
-
             m = self.BACKTRACE_PATTERN.match(strip_line)
             if m is None:
                 continue
--- a/monitor/filter_exception_decoder.py
+++ b/monitor/filter_exception_decoder.py
@@ -34,6 +34,7 @@ class Esp32ExceptionDecoder(DeviceMonitorFilterBase):
 
     BACKTRACE_PATTERN = re.compile(r"^Backtrace:(((\s?0x[0-9a-fA-F]{8}:0x[0-9a-fA-F]{8}))+)")
     BACKTRACE_ADDRESS_PATTERN = re.compile(r'0x[0-9a-fA-F]{8}:0x[0-9a-fA-F]{8}')
+    ANSI_ESCAPE_PATTERN = re.compile(r'(\x9B|\x1B\[)[0-?]*[ -\/]*[@-~]')
 
     def __call__(self):
         self.buffer = ""
@@ -49,6 +50,11 @@ Please build project in debug configuration to get more details about an excepti
 See https://docs.platformio.org/page/projectconf/build_configurations.html
 
 """
+            )
+
+        if self.config.get("env:" + self.environment, "monitor_raw") == "yes":
+            sys.stderr.write(
+                "%s: The filter is disabled and will not be called, due to 'monitor_raw' = 'yes'.\n" % self.__class__.__name__
             )
 
         return self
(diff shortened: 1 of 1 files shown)


BOBAH1248:
refs [#4556 ](https://github.com/platformio/platformio-core/issues/4556)

monitor/filter_exception_decoder.py
```diff
@@ -100,7 +112,12 @@ def rx(self, text):
                 self.buffer = ""
             last = idx + 1
 
-            m = self.BACKTRACE_PATTERN.match(line)
+            strip_line = self.strip_ansi(line)
+            # sys.stderr.write(
```
MacDada, line 116: commented out code?

BOBAH1248: commented lines 116-118?
I left it because I spent time to find way how to see in debug logs non-printable characters

MacDada: Well, I'm not a maintainer here, so I have no saying, but in general – in the places that I do have a saying ;) – I forbid leaving commented out code, most of the time ;)

Better to add a dedicated function for that, with a switch or not, or whatev, to be reusable. Or just remove it.

BOBAH1248: Sounds reasonable
In the previous commit, I removed these lines, but after a while they were needed again.
By the way, have you knowledge how I can add a switch to `platformio.exe device monitor` command to get it in the "addon" script?

MacDada: > how I can add a switch to `platformio.exe device monitor` command to get it in the "addon" script?

Sorry, IDK

valeros:
Hi @BOBAH1248 ! Do you mind sharing a minimal but configured PlatformIO project to evaluate your changes? Also, I see that some code overlaps with #1039? Does this PR supersedes it?

BOBAH1248:
@valeros, It is a difficult task to develop a project like this, where every time the problem was repeated with 100% probability.
With logs, it's mostly a matter of chance. Otherwise, I am sure that you would have discovered and corrected it long ago.

In any case, I can only say with 100% certainty that:
- when `monitor_raw = yes`, the execution of filters from `monitor_filters` does not occur (filters loaded, but not processed)
- `monitor_filters = colorize,...` break the output to Windows terminal
- without filters (when not present `monitor_filters =...`) the colors are not recognized (printed ascii name like `ESC`)
- with just `esp32_exception_decoder` filter (i.e. `monitor_filters = esp32_exception_decoder`), the colors are recognized (but I expected that output should be like in previous row)

This behavior was when I publish the PR request (and I not observed ascii name like `ESC`).
But then delivered Windows and VS Code updates, so now the initial behavior is not the same - something works, something not, something less observed (like [#4556](https://github.com/platformio/platformio-core/issues/4556))

Any way, I attached minimal project with few configurations and comments in project file
[esp32_monitor_filters_test.zip](https://github.com/platformio/platform-espressif32/files/10921213/esp32_monitor_filters_test.zip)

BOBAH1248:
> Does this PR supersedes it?

Yes

But the issue, that fixed by my second PR, is not replicated on the fully updated to last release of Windows system, VS Code and PlatformIO.

valeros:
Unfortunately, I also cannot reproduce the issue with everything updated with or without the code from your PR. I may be missing the point, so let's clear things up:

> when monitor_raw = yes, the execution of filters from monitor_filters does not occur (filters loaded, but not processed)

That's actually expected behavior, so you cannot use the `monitor_raw` and `monitor_filters` options together.

> monitor_filters = colorize,... break the output to Windows terminal
> without filters (when not present monitor_filters =...) the colors are not recognized (printed ascii name like ESC)

I guess the `colorize` filter is not the right option in your case as your output is already colorized on your device, so you only need to let your shell print it correctly. I believe `monitor_filters = direct, esp32_exception_decoder` is the right combination.

Finally, please keep in mind that you always must have the `firmware.elf` file in place, otherwise the `esp32_exception_decoder` decoder will simply return unprocessed output. PlatformIO clears up the build folder after any change to `platformio.ini` so when testing please rebuild the project.

BOBAH1248:
With the latest Windows and VS Code updates the issues does not replicated.

PR #1054, opened 2023-03-16
boards: initial support for ESP32-C6-DevKitC-1

stintel:


Jason2866:
You are not using latest espidf.py version. C6 has no support in Arduino. Needs to be removed from the boards.json

stintel:
Used this to successfully flash an esphome firmware to an ESP32-C6. Removed the unrelated commit from this PR, so I think this is ready for review. One question I have is about maximum_ram_size ... The ESP32-C6 has 512K RAM, but I'm not sure if this should be reflect in maximum_ram_size.  Also the url currently returns a 404, should I remove it for now?

valeros:
Hi @stintel ! Thanks for the PR. 

> The ESP32-C6 has 512K RAM, but I'm not sure if this should be reflect in maximum_ram_size

If all 512K of RAM is available for use, then you can transfer this value to `maximum_ram_size`.

> Also the url currently returns a 404, should I remove it for now?

Isn't this https://docs.espressif.com/projects/espressif-esp-dev-kits/en/latest/esp32c6/esp32-c6-devkitc-1/index.html the right link?

Also, there still some mentions of the Arduino framework in the board manifest, I guess it should be removed until it's officially supported.

linkedupbits:
ESP32-c6 is now supported via ESP-IDF 5.1
https://github.com/espressif/esp-idf/releases/tag/v5.1
What do we need to make this work?

lboue:
Could a maintainer review this PR?

Jason2866:
Use this changes in my fork. Could not test with IDF5.1 since i got it not running (compile does work but get a linker error...). But it does work well with the early alpha Arduino 3.0 build. C6 Arduino Blink example is working.
The PR is fine

nomis:
> > If all 512K of RAM is available for use, then you can transfer this value to `maximum_ram_size`.
> 
> From https://www.espressif.com/sites/default/files/documentation/esp32-c6-wroom-1_datasheet_en.pdf:
> 
> CPU and On-Chip Memory • ESP32-C6 embedded, 32-bit RISC-V single-core microprocessor, up to 160 MHz • ROM: 320 KB • HP SRAM: 512 KB • LP SRAM: 16 KB
> 
> To be honest, I'm not entirely sure if all of it is available for use.

The RAM size should be set to 512KB. There are already other misconfigured ESP32 boards (all of the S3s) set to 320KB because they've just been copied from a 320KB model when it actually has 512KB.

stintel:
> The RAM size should be set to 512KB.

Done.

Jason2866:
@nomis  The ESP32-x RAM settings for the xtensa MCUs are correct the physical existing RAM of 512kb BUT can not be used by the sketch.

valeros:
Many thanks for a great PR, merged!

PR #1067, opened 2023-04-03
Update .json files of Deneyap Boards

DogushC:
Update .json files of Deneyap Boards

### DIFF
--- a/boards/deneyapkart.json
+++ b/boards/deneyapkart.json
@@ -14,7 +14,7 @@
     ],
     "f_cpu": "240000000L",
     "f_flash": "80000000L",
-    "flash_mode": "dio",
+    "flash_mode": "qio",
     "mcu": "esp32",
     "variant": "deneyapkart"
   },
--- a/boards/deneyapkart.json
+++ b/boards/deneyapkart.json
@@ -7,19 +7,20 @@
     "extra_flags": [
       "-DARDUINO_DYDK",
       "-DBOARD_HAS_PSRAM",
+      "-DARDUINO_RUNNING_CORE=1",
+      "-DARDUINO_EVENT_RUNNING_CORE=1",
       "-mfix-esp32-psram-cache-issue",
       "-mfix-esp32-psram-cache-strategy=memw"
     ],
     "f_cpu": "240000000L",
     "f_flash": "80000000L",
-    "flash_mode": "qio",
+    "flash_mode": "dio",
     "mcu": "esp32",
     "variant": "deneyapkart"
   },
   "connectivity": [
     "wifi",
-    "bluetooth",
-    "ethernet"
+    "bluetooth"
   ],
   "debug": {
     "openocd_board": "esp32-wrover.cfg"
@@ -36,6 +37,6 @@
     "require_upload_port": true,
     "speed": 460800
   },
-  "url": "https://deneyapkart.org/magaza/urun-deneyap-kart.html",
-  "vendor": "Deneyap"
+  "url": "https://magaza.deneyapkart.org/tr/product/detail/deneyap-kart",
+  "vendor": "T3 Foundation"
 }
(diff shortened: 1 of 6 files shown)


boards/deneyapkart.json
```diff
@@ -7,19 +7,20 @@
...
       "-DBOARD_HAS_PSRAM",
+      "-DARDUINO_RUNNING_CORE=1",
+      "-DARDUINO_EVENT_RUNNING_CORE=1",
       "-mfix-esp32-psram-cache-issue",
       "-mfix-esp32-psram-cache-strategy=memw"
     ],
     "f_cpu": "240000000L",
     "f_flash": "80000000L",
-    "flash_mode": "qio",
+    "flash_mode": "dio",
```
valeros, line 17: Could you please explain the reasoning behind changing the `flash_mode` option? It seems `QIO` is used when selecting bootloader for the Deneyap boards in the Arduino core, is it wrong here https://github.com/espressif/arduino-esp32/blob/master/boards.txt#L17358 ?

boards/deneyapmini.json
```diff
@@ -37,11 +35,9 @@
     "flash_size": "4MB",
     "maximum_ram_size": 327680,
     "maximum_size": 4194304,
-    "use_1200bps_touch": true,
-    "wait_for_upload_port": true,
```
valeros, line 41: https://github.com/espressif/arduino-esp32/blob/master/boards.txt#L17979
boards.txt states that this board requires 1200bps touch by default, is it wrong?

valeros (changes requested):

valeros:
Hi @DogushC , thanks for the PR. Please review comments above.

boards/deneyapkart.json
DogushC: > Could you please explain the reasoning behind changing the `flash_mode` option? It seems `QIO` is used when selecting bootloader for the Deneyap boards in the Arduino core, is it wrong here https://github.com/espressif/arduino-esp32/blob/master/boards.txt#L17358 ?

I think that "flash_mode" parameter is equivalent at https://github.com/espressif/arduino-esp32/blob/master/boards.txt#L17357. 
If it is the equivalent of the "boot" parameter of arduino-esp32 repo, I have to update it here.

boards/deneyapmini.json
DogushC: > https://github.com/espressif/arduino-esp32/blob/master/boards.txt#L17979 boards.txt states that this board requires 1200bps touch by default, is it wrong?

You are right. I will update it.

boards/deneyapkart.json
valeros: > If it is the equivalent of the "boot" parameter of arduino-esp32 repo, I have to update it here.

Yes, it's equivalent, so I'd recommend using bootmode value from boards.txt in this field. The transformation to a proper `flash_mode` happens here https://github.com/platformio/platform-espressif32/blob/develop/builder/main.py#L68

boards/deneyapmini.json
valeros: > You are right. I will update it.

The same with `deneyapminiv2`, seems like 1200bps is required.

boards/deneyapkart.json
DogushC: Update all `flash_mode` parameter.

boards/deneyapmini.json
DogushC: Fix `use_1200bps_touch` and `wait_for_upload_port` parameters of Deneyap Mini v2, Deneyap Mini and Deneyap Kart 1A v2.

DogushC: We have a [PR](https://github.com/espressif/arduino-esp32/pull/8046) for update `use_1200bps_touch` and `wait_for_upload_port` for Deneyap Kart 1A v2.

valeros:
@DogushC do you plan any further updates?

DogushC:
> @DogushC do you plan any further updates?

No. Thanks.

valeros:
Thanks, merged.

PR #1082, opened 2023-04-22
Fix Seeed XIAO ESP32S3 PSRAM

alwint3r:
I tried running a code that uses the PSRAM, but I got the following error:

```
E (164) psram: PSRAM ID read error: 0x00ffffff
```

I changed the `memory_type` of the seeed_xiao_esp32s3.json definition file to `qio_opi` and then the PSRAM issue is fixed.

This PR intends to fix this problem.

### DIFF
--- a/boards/seeed_xiao_esp32s3.json
+++ b/boards/seeed_xiao_esp32s3.json
@@ -3,7 +3,7 @@
     "arduino": {
       "ldscript": "esp32s3_out.ld",
       "partitions": "default_8MB.csv",
-      "memory_type": "qio_qspi"
+      "memory_type": "qio_opi"
     },
     "core": "esp32",
     "extra_flags": [


valeros:
Hi @alwint3r, thanks for the PR. For some reason PSRAM is [disabled](https://github.com/espressif/arduino-esp32/blob/master/boards.txt#L20330) in the Arduino core. Is it wrong there?

Jason2866:
It is wrong in Arduino too. The used module has OPI flash.

alwint3r:
Hi, @valeros and @Jason2866. The PSRAM is disabled by default if you are using Arduino IDE. It can be enabled by choosing the OPI PSRAM option. So, I don't think it's wrong. Maybe people from Seeed Studio decided that the PSRAM is disabled by default for Arduino IDE users.

PR #1139, opened 2023-07-01
Fix OpenOCD board -> target on Lolin S2 Mini

maxgerhardt:
There is no `board/esp32s2.cfg`. But there is `target/esp32s2.cfg`.

Reported per [community forum](https://community.platformio.org/t/error-cant-find-board-esp32s2-cfg-in-procedure-script/34544).


valeros:
Many thanks!

PR #1176, opened 2023-08-26
bee_motion_s3: use s3 linker script

Jason2866:


valeros:
Thanks for the PR, merged.

PR #1181, opened 2023-09-03
Add Adafruit QT PY S3 with 4MB flash 2 MB PSRAM

R41K:
Please add support for this board: https://www.adafruit.com/product/5700 .
The _nopsram variant does not work because it requires 8MB flash_size but this board has only 4mb flash + 2mb psram.
I tested it with a blink sketch.

### DIFF
--- a/boards/adafruit_qtpy_esp32s3_n4r2.json
+++ b/boards/adafruit_qtpy_esp32s3_n4r2.json
@@ -1,0 +1,49 @@
+{
+  "build": {
+    "arduino":{
+      "ldscript": "esp32s3_out.ld",
+      "partitions": "partitions-4MB-tinyuf2.csv"
+    },
+    "core": "esp32",
+    "extra_flags": [
+      "-DARDUINO_ADAFRUIT_QTPY_ESP32S3_N4R2",
+      "-DARDUINO_USB_CDC_ON_BOOT=1",
+      "-DARDUINO_RUNNING_CORE=1",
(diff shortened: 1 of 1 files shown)


awwaiid:
I'm using the slightly different variant at https://community.platformio.org/t/adafruit-qt-py-s3-with-2mb-psram/37256/2 and it works great.

R41K:
Hi @valeros, I am not much of a developer and not so deep in PIO but I gave it a shot and created this PR. Apparently it was kind of ignored/unprocessed for three months without any obvious reason. So may I ask if anything was wrong with this PR? It makes me wonder a bit why #1257 got such a fast response and got pulled almost immediately. Even though there was no support of TinyUF2 and some other small things, I would have expected that these topics come up when there is a Review from your side, which did not happen. I've spent some time and effort on this and want to avoid other experiences like this in the future. Would be nice to get some insights why it was unprocessed for so long and how to get requests processed faster in the future. Thanks, R41K

valeros:
Hi @R41K, I'm sorry that your PR didn't get much attention, please be sure I'm not prejudiced against anyone. Usually, I review PRs with new board manifests when I'm updating the Arduino core to the latest version. This last update of the Arduino core with support for the Adafruit QT PY S3 board was significantly delayed (as you mentioned for several months), so that when the moment to update came, there was that new PR with a bit better implementation. 

All contributions are welcome, if the new PR with support for TUF2 hadn't been added, of course I would've reviewed yours. It was just a unfortunate coincidence that your PR has gone unnoticed.

PR #1198, opened 2023-09-21
Update lolin_s3.json

f-io:
Board was missing BT connectivity. 
PSRAM type is octal SPI => opi.

https://www.wemos.cc/en/latest/s3/s3.html

### DIFF
--- a/boards/lolin_s3.json
+++ b/boards/lolin_s3.json
@@ -15,6 +15,7 @@
     "f_cpu": "240000000L",
     "f_flash": "80000000L",
     "flash_mode": "qio",
+    "psram_type": "opi",
     "hwids": [
       [
         "0x303A",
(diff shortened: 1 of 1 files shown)


boards/lolin_s3.json
```diff
@@ -15,6 +15,7 @@
     "f_cpu": "240000000L",
     "f_flash": "80000000L",
     "flash_mode": "qio",
+    "psram_type": "opi",
```
valeros, line 18: Isn't the PSRAM type already covered in just above in the memory type field? Where do you use it?
```json
{
  "build": {
    "arduino": {
      "ldscript": "esp32s3_out.ld",
      "partitions": "default_16MB.csv",
      "memory_type": "qio_opi"
    },
...
}
```

f-io: My fault, did not look into /builder/main.py (def _get_board_memory_type(env)), removed duplicate.

valeros:
Thanks for the PR, merged!

PR #1200, opened 2023-09-24
Add ftSwarm[Rs/Control] Board

Bloeckchengrafik:


valeros:
Hi @Bloeckchengrafik, it seems there are no `ftswarm` and `ftswarm_rs` variants in the official Arduino core fore ESP32, so you need to add them to this repo https://github.com/espressif/arduino-esp32 in the first place.

PR #1249, opened 2023-12-09
Added bee data logger board

strid3r21:
hope im doing this right. adding one of my boards for my company Smart Bee. 

my other boards have already been added by someone else at some point, but this particular board is not available in the platformIO IDE yet so im adding it here. (hopefully this is the correct repo to add to so that this option appears in the platformIO IDE.)


valeros:
Thanks for the PR, merged!

PR #1251, opened 2023-12-11
Fix Lolin S2 USB CDC setup

SuGlider:
This commit fixes an issue with this board pointed out in ESP32 Arduino Github.

https://github.com/espressif/arduino-esp32/issues/8977


valeros:
Thanks for the PR, merged.

TheCutter:
This change stopped my project from working using an S2 Mini with USB Serial for logging and Serial0 for external communication. Everything worked with the previous version.

Issue #947, opened 2022-10-28
ESP32-PICO-V3-02 with 8MB partition table is not booting


slaesh:
I am having troubles using the ESP32-PICO-V3-02 with the full flash size of 8MB.
Using the "default.csv" is working fine, but when switching to the "default_8mb.csv" it does not boot any more.. :(

I already had a short chat with @maxgerhardt (thanks again!) [here](https://community.platformio.org/t/support-for-eps32-pico-v3/26967/23?u=slaesh) and he suggested me to file an issue here :)

# repo to reproduce

https://github.com/slaesh/eps32-pico-v3-8mb-flash-issue

# logs using platformio

using platformio and the default partition table, a reset looks like this

```
huhu
huhu
huhu
ets Jul 29 2019 12:21:46

rst:0x1 (POWERON_RESET),boot:0x13 (SPI_FAST_FLASH_BOOT)
configsip: 271414342, SPIWP:0xee
clk_drv:0x00,q_drv:0x00,d_drv:0x00,cs0_drv:0x00,hd_drv:0x00,wp_drv:0x00
mode:DIO, clock div:2
load:0x3fff0030,len:1344
load:0x40078000,len:13836
load:0x40080400,len:3608
entry 0x400805f0
[     4][D][esp32-hal-cpu.c:244] setCpuFrequencyMhz(): PLL: 480 / 2 = 240 Mhz, APB: 80000000 Hz
[   228][I][esp32-hal-psram.c:96] psramInit(): PSRAM enabled
huhu
huhu
huhu
huhu
```

using platformio and the 8mb partition table, this error loops around:

```
ets Jul 29 2019 12:21:46

rst:0x3 (SW_RESET),boot:0x13 (SPI_FAST_FLASH_BOOT)
configsip: 271414342, SPIWP:0xee
clk_drv:0x00,q_drv:0x00,d_drv:0x00,cs0_drv:0x00,hd_drv:0x00,wp_drv:0x00
mode:DIO, clock div:2
load:0x3fff0030,len:1344
load:0x40078000,len:13836
load:0x40080400,len:3608
entry 0x400805f0
ets Jul 29 2019 12:21:46

rst:0x3 (SW_RESET),boot:0x13 (SPI_FAST_FLASH_BOOT)
configsip: 271414342, SPIWP:0xee
clk_drv:0x00,q_drv:0x00,d_drv:0x00,cs0_drv:0x00,hd_drv:0x00,wp_drv:0x00
[...]

PR #1254, opened 2023-12-14
Added board ESP32-PICO-DevKitM-2.

djpearman:
Hi,

Added board definition file for [ESP32-PICO-DevKitM-2](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/hw-reference/esp32/get-started-pico-devkitm-2.html).

Best regards,
djpearman

### DIFF
--- a/boards/esp32-pico-devkitm-2.json
+++ b/boards/esp32-pico-devkitm-2.json
@@ -1,0 +1,37 @@
+{
+  "build": {
+    "arduino": {
+      "ldscript": "esp32_out.ld"
+    },
+    "core": "esp32",
+    "extra_flags": [
+      "-DARDUINO_ESP32_PICO_DEVKITM_2",
+      "-DBOARD_HAS_PSRAM"
+    ],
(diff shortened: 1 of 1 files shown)


valeros:
Hi @djpearman, thanks for the PR. Aren't `-mfix-esp32-psram-cache-issue` and `-mfix-esp32-psram-cache-strategy=memw` flags required for proper PSRAM support on ESP32?

Jason2866:
@valeros The options are only needed for early esp32 board (to be exact rev 0 and 1)
Since the pico variant is a later rev. it is not needed.

djpearman:
Hi @valeros,

I had those two flags in there initially, but came across [a comment](https://github.com/platformio/platform-espressif32/issues/947#issuecomment-1295306307) stating that the PICO-V3-02 no longer requires them. Removed them, tested it and it works. ESP.getPsramSize() and ESP.getFreePsram() report values, so I figure it works.

valeros:
Many thanks, merged!

Issue #1211, opened 2023-10-18
Track Arduino ESP32 Alpha Release v3.0.0 based on ESP-IDF v5.1 Support Arduino ESP32 v3.0 based on ESP-IDF v5.1


brentru:
@maxgerhardt  @ivankravets  Hi, I was wondering if there are any plans to synchronize the development branch of this repository with the IDF v5.1 alpha releases? I understand that these releases are in the alpha stage, but I believe that updating the PIO package to be compatible with IDF v5+ could greatly benefit the community.

If this isn't in the roadmap, could you please suggest any potential workarounds?

[Arduino Alpha 2 Release v3.0.0 based on ESP-IDF v5.1](https://github.com/espressif/arduino-esp32/releases/tag/3.0.0-alpha2)

maxgerhardt:
Can't speak for PIO Labs as a company since I'm not even an employee, but I've read multiple times that PIO's policy is to support stable version, not alpha / beta version. So if it's released as stable, support will probably follow.

Nevertheless, @Jason2866 has a `Arduino/IDF5` branch in his https://github.com/Jason2866/platform-espressif32. So your first try should be to just use
```
platform = https://github.com/Jason2866/platform-espressif32.git#Arduino/IDF5
```
in your `platformio.ini` and seee where that takes you.

mhaberler:
@maxgerhardt thanks!

I tried your suggestion with a small esp32s3 project with Arduino as an esp-idf component using PlatformIO 6.1.11

>platform = https://github.com/Jason2866/platform-espressif32.git#Arduino/IDF5
framework = arduino, espidf

build and running works fine

what does not work any more is JTAG debugging as gdb is not found
> FileNotFoundError: [Errno 2] No such file or directory: 'xtensa-esp32s3-elf-gdb'

I guess this has to do with the factoring out of esp32 gdb into its own repo, it is not present in .platformio/packages/toolchain-xtensa-esp32s3/bin/

I assume this would be available from https://docs.espressif.com/projects/esp-idf/en/v5.1.1/esp32s3/api-guides/tools/idf-tools.html?highlight=xtensa%20esp32s3%20elf%20gdb#xtensa-esp-elf-gdb 

but I'm at loss on how to add this to my installation/project?

Jason2866:
The `gdb` are listed in the `platform.json`. I will have a look why the are not installed.
My Python Build Scripts differ from the official ones. Probably overlooked to install them...

[...]

PR #1281, opened 2024-01-14
Support Arduino ESP32 alpha v3.0.0 based on ESP-IDF v5.1

sgryphon:
PlatformIO already supports overriding platform packages to use an alternative or development version of a package, however a missing library dependency means this doesn't work for arduino-esp32 alpha v3.0.0.

This PR adds a reference to framework-arduinoespressif32-libs SDK (aka esp32-arduino-libs), so that the latest arduino-esp32 alpha can be used. But only if you override platform packages to the alpha 3.0.0 version, which references libs, and also point at the libs location.

The change is backwards compatible, i.e. does not break any existing code, as the library framework-arduinoespressif32-libs is only included if it is referenced in an overridden arduino-esp32 manifest.

This fixes issue #1225 and #1211, i.e. if you are getting the error "KeyError: 'framework-arduinoespressif32-libs':" due to the missing library (when you reference an alpha v3 version of arduino-esp32)

To use, configure your project to reference the latest arduino-esp32, which will include a reference to framework-arduinoespressif32-libs. You also need to override the location of the libs (as they are not currently in the PlatformIO repository).

```
platform = https://github.com/sgryphon/platform-espressif32.git#sgryphon/add-esp32-arduino-libs
platform_packages =
    platformio/framework-arduinoespressif32 @ https://github.com/espressif/arduino-esp32.git#master
    platformio/framework-arduinoespressif32-libs @ https://github.com/espressif/esp32-arduino-libs.git#idf-release/v5.1
```

If you use a older version of the platform package, e.g. the default, then libs are not needed, and won't be loaded. So the following also works fine (old version of platform package, no libs), or any override of the esp32-arduino platform package that does not include the libs (i.e. pre alpha v3) will also work:

```
platform = https://github.com/sgryphon/platform-espressif32.git#sgryphon/add-esp32-arduino-libs
platform_packages =
```

This change simply allows the arduino-esp32 platform package to be configured with alpha 3.0.0, if you want to test some advanced development (without affecting existing support).

Future work: The library dependency needs to be added to the PlatformIO Registry, and when 3.0.0 is released/stable, the dependency can be made Optional = False as the default.

**Note:** This change doesn't necessarily mean your code will work, only that the missing dependency will be there. Alpha 3.0.0 has breaking changes, so existing code/libraries may not be compatible. The esp32-arduino package does have some basic examples that do work, and are run as part of their integrations tests. https://github.com/espressif/arduino-esp32/blob/e4d6a8abf9fab7422e21f72d3aeb2e9c9484ab36/.github/scripts/on-push.sh#L92

### DIFF
--- a/platform.json
+++ b/platform.json
@@ -81,6 +81,12 @@
       "owner": "platformio",
       "version": "~3.20014.0"
     },
+    "framework-arduinoespressif32-libs": {
+      "type": "sdk",
+      "optional": true,
+      "owner": "espressif",
+      "version": "~5.1.0"
+    },
     "framework-arduino-mbcwb": {
       "type": "framework",
       "optional": true,
--- a/platform.py
+++ b/platform.py
@@ -342,6 +342,7 @@ class Espressif32Platform(PlatformBase):
             )
 
         toolchain_remap = {
+            "esp32-arduino-libs": "framework-arduinoespressif32-libs",
             "xtensa-esp32-elf-gcc": "toolchain-xtensa-esp32",
             "xtensa-esp32s2-elf-gcc": "toolchain-xtensa-esp32s2",
             "xtensa-esp32s3-elf-gcc": "toolchain-xtensa-esp32s3",
@@ -401,6 +402,8 @@ class Espressif32Platform(PlatformBase):
             self.packages[toolchain_package]["version"] = version
             self.packages[toolchain_package]["owner"] = "espressif"
             self.packages[toolchain_package]["type"] = "toolchain"
+            if (toolchain_package == "framework-arduinoespressif32-libs"):
+                self.packages[toolchain_package]["optional"] = False
 
     def configure_upstream_arduino_packages(self, url_items):
         framework_index_file = os.path.join(
--- a/platform.py
+++ b/platform.py
@@ -321,6 +321,7 @@ class Espressif32Platform(PlatformBase):
                 r"^gcc(?P<MAJOR>\d+)_(?P<MINOR>\d+)_(?P<PATCH>\d+)-esp-(?P<EXTRA>.+)$",
                 r"^esp-(?P<EXTRA>.+)-(?P<MAJOR>\d+)\.(?P<MINOR>\d+)\.?(?P<PATCH>\d+)$",
                 r"^esp-(?P<MAJOR>\d+)\.(?P<MINOR>\d+)\.(?P<PATCH>\d+)(_(?P<EXTRA>.+))?$",
+                r"^idf-release_v(?P<MAJOR>\d+)\.(?P<MINOR>\d+)(.(?P<PATCH>\d+))?(-(?P<EXTRA>.+))?$",
             )
             for pattern in version_patterns:
                 match = re.search(pattern, original_version)
@@ -328,7 +329,7 @@ class Espressif32Platform(PlatformBase):
                     result = "%s.%s.%s" % (
                         match.group("MAJOR"),
                         match.group("MINOR"),
-                        match.group("PATCH"),
+                        match.group("PATCH") if match.group("PATCH") is not None else "0",
                     )
                     if match.group("EXTRA"):
                         result = result + "+%s" % match.group("EXTRA")


philm001:
Does this work for all ESP32 boards? Or just certain ones

sgryphon:
> Does this work for all ESP32 boards? Or just certain ones

This change isn't directly related to boards, but to PlatformIO setup. There is no board specific stuff.

All it will do is allow you to get past the error "KeyError: 'framework-arduinoespressif32-libs':", when setting up the platform, even before compiling anything.

Once past that error, you application will probably fail to compile as there are many breaking changes in Arduino ESP32 alpha v3 (and it is alpha).

e.g. Even a very basic example on my M5Core2 failed -- neither the M5Core2 or M5Unified libraries would even compile

However, removing the include reference to any M5 stuff, and just using core Arduino (e.g. start serial and output a counter), does build and run, so core platform stuff works.

This change to allow PlatformIO to at least set up the dependencies, will then allow peeps to test whether their libraries, etc will work with v3.0.0 and update them as necessary.

There is a migration guide https://docs.espressif.com/projects/arduino-esp32/en/latest/migration_guides/2.x_to_3.0.html

Note that the new libs will only be loaded if you override platform packages for arduino-esp32 to point to an alpha 3 version (that references it).

sgryphon:
> Does this work for all ESP32 boards? Or just certain ones

As mentioned, this change (allowing the PlatformIO integration to set up the libs dependency) doesn't have anything directly to do with boards ... but what might be relevant is what the actual library supports.

The libs referenced in the example configuration is for ESP-IDF v5.1 from https://github.com/espressif/esp32-arduino-libs

Inside that ZIP file are folders for esp32, esp32c3, esp32c6, esp32h2, esp32s2, and esp32s3

Note that this change (in the setup scripts) doesn't directly reference this library ... all the change does is allow you to add the reference to your platform.io config (well, it does define a default of 5.1.0, but the library hasn't be added to the PlatformIO registry so that won't do anything anyway -- which is why it is optional and only configured if you reference an alpha platform that includes it).

You could also configure libs to point at a custom build, e.g. something based on ESP-IDF 5.2 or 5.3.

This change is about giving you the ability to make that reference, not what is necessarily what is in the reference. As mentioned alpha 3.0.0 has a bunch of breaking changes, so your app might not work or even compile.

Jason2866:
For C6 and H2 custom boards.json are needed. The existing ones (if there are any) don't have the entry to support Arduino as framework.

sgryphon:
> Sure wish I knew what was holding back 3.0.0/IDF 5.1.x adoption in platformIO, thank you VERY much for figuring this setup out.

Probably the main reason is that 3.0.0 is only RC1, and not officially released yet.

PlatformIO does allow overriding of the URLs, so you can usually easily point to an updated / RC version (or even your own custom one). That has always been supported.

In this case it is a bit different, because the base script needs to account for a new libs component (what this PR does).

They could merge this PR now; it doesn't break anything existing, and enables support for the alpha/RC version. But it is also reasonable for them not to merge in something that is only for an unreleased thing (e.g. what happens if it changes, e.g. 3 components or merges back into 1).

thijstriemstra:
> Probably the main reason is that 3.0.0 is only RC1, and not officially released yet.

I highly doubt this. #1225 clearly states espressif dropped platformio starting from 3.0. We'll see...

songpola:
> > Probably the main reason is that 3.0.0 is only RC1, and not officially released yet.
> 
> I highly doubt this. #1225 clearly states espressif dropped platformio starting from 3.0. We'll see...

They updated that they didn't drop PlatformIO support entirely.
They just stated that the PlatformIO supports are not maintained by them officially but are maintained by the PlatformIO community.

sgryphon:
> #1225 clearly states espressif dropped platformio starting from 3.0. We'll see...

#1225 is a post by Platform IO ... yes, it claims that Espressif has discontinued support, but Espressif has said no such thing. If you follow the link to the "official statement" it is a pull request updating their documentation.

Initially it removed details about PlatformIO from the Espressif documentation because it confused users (Espressif has no control over PlatformIO or anyway to provide support for it); if you follow down the PR, that was then changed (before merging) to keep the details and instead just having a warning that PlatformIO issues need to be reported to PlatformIO (not Espressif).

You can see the current (RC1) version of the page: https://github.com/espressif/arduino-esp32/blob/3.0.0-rc1/docs/en/installing.rst#installing-using-platformio

No where is there any mention of dropping Platform IO -- it's still right there.

sgryphon:
#master is volatile, so could end up with issues. It was working around the time of #3.0.0-rc1 (about 30 commits / 3 weeks ago).

Which version of arduino-esp32 and the libs you use may vary -- the platform script changes are simply to allow you to make the references.

I will have a look to see if it is something about the current master, the platform scripts (may need rebasing), or another issue, e.g. changes in alpha/rc 3.0.0 may simply not be compatible.

Jason2866:
It is a bug here https://github.com/espressif/arduino-esp32/blob/5a023568a640f7b55f40381a28389d05df3af9c5/package/package_esp32_index.template.json#L45
there is the branch missing

connornishijima:
In case it helps anyone, here is my current platformio.ini that successfully compiles Arduino 3.0.0-rc1 (with IDF v5.1.3-470-g2fd087b246-dirty behind the scenes)

All I had to do was manually make [the changes defined in sgryphon's PR](https://github.com/platformio/platform-espressif32/pull/1281/files) to my platformio files.

    platform = https://github.com/platformio/platform-espressif32.git#develop
    platform_packages = 
        platformio/framework-arduinoespressif32 @ https://github.com/espressif/arduino-esp32.git#3.0.0-rc1
        platformio/framework-arduinoespressif32-libs @ https://github.com/espressif/esp32-arduino-libs.git#idf-release/v5.1

https://github.com/Lixie-Labs/Emotiscope/blob/1.1/platformio.ini

Jason2866:
@connornishijima Yep, that's the simplest and easiest solution.

Hedda:
FYI, Espressif released "stable" version 3.0.0 Arduino ESP32 Core based on ESP-IDF v5.1.4 with support for ESP32-H2 and ESP32-C6:

https://github.com/espressif/arduino-esp32/releases/tag/3.0.0

Official documentation in arduino-esp32 v3.0.0 from Espressif still have instructions for installing using PlatformIO:

* https://github.com/espressif/arduino-esp32/blob/3.0.0/docs/en/installing.rst#installing-using-platformio

(Arduino ESP32 core v3.0.0 was first announced over 7-months ago but this updated version is now finally been marked as stable).

- https://github.com/platformio/platform-espressif32/issues/1225

PS: Also means can report non-PlatformIO problems and feature requests related to upstream ESP32-H2/ESP32-C6 as new issues:

https://github.com/espressif/arduino-esp32/issues/

Jason2866:
@sgryphon You need to do changes in your PR to support actual espressif32 Arduino core 3.1.0
BTW. You can close the PR. It will not be merged anytime. There will be no support for core > 2.x. from Platformio

thijstriemstra:
>  There will be no support for core > 2.x. from Platformio

Time to fork it?

Jason2866:
@thijstriemstra it is already forked

thijstriemstra:
I didn't know, thanks for the headsup

https://www.cnx-software.com/2024/08/27/platform-espressif32-fork-platformio-arduino-esp32-c6-esp32-c5-esp32-h2-esp32-p4/

djfm:
I might not be aware of all the issues, but after playing around with arduino, arduino-esp32 in PlatformIO and ESP-IDF standalone (which also has a pretty decent vscode extension), my intuition would be that it should be easier to make esp-idf API-compatible with Arduino through a thin wrapper included as an ESP-IDF component than making Arduino able to call into esp-idf's API's, which is, I think, what arduino-espressif32 is attempting to do, but with a lot of downsides (compatibility between latest versions is not trivial, `no idf.py menuconfig`)...

Unfortunately, being in a production rush I have had very little time to look into this and maybe I'm just saying dumb shit, feel free to correct me if I'm wrong.

Anyways the most expedient strategy for me has been to go full ESP-IDF, but it would be nice to be able to use some Arduino libs. My gut feeling is that an ESP-IDF compatibility component alllwing to use Arduino libs de esp-idf shouldn't be that complicated to do...

PR #1298, opened 2024-01-30
Add board ESP32-S3-Devkitc-1-N8R8 variant

lranders:
Seeing how many people struggle to enable PSRAM, perhaps a working template would be helpful.

It's a carbon copy of the vanilla `esp32-s3-devkitc-1.json` file with the required modifications.

### DIFF
--- a/boards/esp32-s3-devkitc-1-n8r8.json
+++ b/boards/esp32-s3-devkitc-1-n8r8.json
@@ -1,0 +1,53 @@
+{
+  "build": {
+    "arduino":{
+      "ldscript": "esp32s3_out.ld",
+      "memory_type": "qio_opi",
+      "partitions": "default_8MB.csv"
+    },
+    "core": "esp32",
+    "extra_flags": [
+      "-DBOARD_HAS_PSRAM",
+      "-DARDUINO_ESP32S3_DEV",
+      "-DARDUINO_USB_MODE=1",
+      "-DARDUINO_RUNNING_CORE=1",
+      "-DARDUINO_EVENT_RUNNING_CORE=1"
+    ],
+    "f_cpu": "240000000L",
+    "f_flash": "80000000L",
+    "flash_mode": "qio",
+    "hwids": [
+      [
+        "0x303A",
+        "0x1001"
+      ]
+    ],
+    "mcu": "esp32s3",
+    "variant": "esp32s3"
+  },
+  "connectivity": [
+    "bluetooth",
+    "wifi"
+  ],
+  "debug": {
+    "default_tool": "esp-builtin",
+    "onboard_tools": [
+      "esp-builtin"
+    ],
+    "openocd_target": "esp32s3.cfg"
+  },
+  "frameworks": [
+    "arduino",
+    "espidf"
+  ],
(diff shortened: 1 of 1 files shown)


Jason2866:
Fight against wind mills by adding another variant of the many different s3 devkits.
And it is wrong, when connecting via the UART board and not the assumed CDC port.

The boards concept as used now is a dead end. Better way would to list the features in name.

lranders:
I am not entirely sure what to make of those comments @Jason2866...

Is there anything I can do to fix your obvious objections or did I inadvertently wander into a minefield here?
Am I understanding you correctly that you do not want any more board submissions?

Jason2866:
@lranders Just a note that adding more and more boards will not solve the situation that without an usefull info in the boards manifest name it is just another one.

In this case where can it be seen it is for using the CDC port?

lranders:
> Is it useful to have a boards list with a few hundred boards? Basically all the same just with an other name.

Considering how many stumble, just as I did, before being put on the right track, actually I do.
You're right; everything _could_ be defined in PlatformIO setup .ini and then we could do away with board files entirely,
but it does strike me as somewhat counterproductive and error-prone compared to the current solution.

I really didn't come here to pick a fight, but it's apparent to me that I stepped right into one of your pet peeves. Sorry!

As for your comment about the CDC define, I amended my original PR comment, stating that I copied it verbatim
from the existing repository file, so maybe a PR for it's removal there would be more appropriate?

Jason2866:
@lranders Sorry, for being personal unfriendly. Was not ment this way. But yes this boards garbage is something i do really dislike.

If we look at this. How does the user know that it is working fully only when connected via the CDC port?

lranders:
Yes, I think you're right about the board needing two different .json files if I understand you correctly.

We're talking about this define `-DARDUINO_USB_MODE=1`, right?
I have no quarrel with removing it if that makes the situation better. Truth be told, as I stated earlier, I have no idea what it does, I just copied the existing file and went with the _lets change as little as possible to make PSRAM work_ approach.

I am still just poking around using the Serial port; haven't gotten around to the USB port yet.

On the other hand, it _is_ possible that I _am_ misunderstanding you and there's no way you'd take this PR.
In that case, how and where could we document what to add to platformio.ini to get PSRAM to work?

Jason2866:
> We're talking about this define -DARDUINO_USB_MODE=1, right?

Yes, it is needed when connected the device via CDC to get serial working. For connecting the device via UART-SERIAL this entry has to be removed. As i mentioned earlier the board needs two boards.json One for connecting via CDC and one for connecting via serial UART

PSRAM is very special with the S3. It depends on which PSRAM type is used (OPI or QIO) and the setting is again different when using OPI as Flash.

By saying this the `esp32-s3-devkitc-1` exists in all possible combinations of Flash and PSRAM types. To get it working the exact matching settings needs to be set. Even when flash size is wrong the device won't boot.

So imho the only way to get out of this mess, is to provide a board configurator where all the relevant settings are asked and the boards.json is generated from this.

lranders:
I finally see your point, but I never imagined it would be this bad.

I totally agree that a board configurator in some shape or form would be far superior to piling on board files
that does or doesn't work, even when one thinks it should.

I apologize for adding to the noise and withdraw my PR.

PR #1308, opened 2024-02-13
Implement `board_img_freq`

Jason2866:
needed since the build recipe in Arduino has changed. https://github.com/espressif/arduino-esp32/blob/master/platform.txt#L159

### DIFF
--- a/builder/main.py
+++ b/builder/main.py
@@ -59,6 +59,15 @@ def _get_board_memory_type(env):
     )
 
 
+def _get_board_img_freq(env):
+    board_config = env.BoardConfig()
+    img_freq = board_config.get("build.img_freq", "")
+    if img_freq =="":
+        img_freq = board_config.get("build.f_flash", "")
+    img_freq = str(img_freq).replace("L", "")
+    return str(int(int(img_freq) / 1000000)) + "m"
+
+
 def _get_board_f_flash(env):
     frequency = env.subst("$BOARD_F_FLASH")
     frequency = str(frequency).replace("L", "")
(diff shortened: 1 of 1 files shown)


valeros:
Hi @Jason2866, thanks for the PR. I guess this new option still defaults to the flash frequency? Does it make sense to reuse already available functionality as a fallback value? Something like this:
```python
def _get_board_img_freq(env):
    return env.BoardConfig().get("build.img_freq", _get_board_f_flash(env))
```

Jason2866:
@valeros yes, the suggestion from you should do too.

valeros:
> Isn't the conversion missing in your suggestion?

Maybe I'm missing something, what conversion do you mean? I expect the end user (or a board manifest) will set the image frequency in the correct format, e.g.:
```ini
platform = espressif32
board = ...
board_build.img_freq = 48m; instead of 48000000L
```

Jason2866:
This is inconsistent since for flash frequency and cpu frequency it is done like this:
```ini
    "f_cpu": "240000000L",
    "f_flash": "80000000L",
```
in boards.json

valeros:
I see, let's then refactor the code a bit, something like this should work:
```python
def _normalize_frequency(frequency):
    frequency = str(frequency).replace("L", "")
    return str(int(int(frequency) / 1000000)) + "m"


def _get_board_f_flash(env):
    frequency = env.subst("$BOARD_F_FLASH")
    return _normalize_frequency(frequency)


def _get_board_img_freq(env):
    board_config = env.BoardConfig()
    if "build.img_freq" in board_config:
        return _normalize_frequency(board_config.get("build.img_freq"))

    return _get_board_f_flash(env)
```

Also, for the sake of consistency with `f_cpu` and `f_flash`, does it make sense to rename this new option to `f_image`?

Jason2866:
Perfect. `f_image` is better since it is more consist with the other options.
Changes done with last commit.

Jason2866:
@valeros Could you merge? Change is full backwards compatible

valeros:
Thanks for the PR, merged.

PR #1315, opened 2024-02-21
Updated some UM boards info 

UnexpectedMaker:
Also renamed tinypico to um_tinypico to match all of the UM board naming. These have been named this way in Arduino for a long time now.

### DIFF
--- a/boards/heltec_wifi_kit_32_V3.json
+++ b/boards/heltec_wifi_kit_32_V3.json
@@ -21,7 +21,7 @@
       ]
     ],
     "mcu": "esp32s3",
-    "variant": "heltec_wifi_kit_32_v3"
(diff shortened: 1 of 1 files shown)


valeros:
Hi @UnexpectedMaker, thanks for the PR. The name of the manifest file plays an important role as a board ID in project configurations. It means that your simple renaming `tinypico.json` to `um_tinypico.json` may break already configured projects. If you insist on renaming, I'd recommend keeping both files `tinypico.json` an `um_tinypico.json`, but the name of the board in the `tinypico.json` file may contain a phrase like `Unexpected Maker TinyPICO (deprecated in favor of "um_tinypico")`, so that it can be safely removed in a new major release.

PR #1317, opened 2024-02-22
Add support for Lilka (v2) board

and3rson:
This PR adds a new board - `Lilka v2`.

Lilka is an open-source community-driven DIY project based on `ESP32-S3-WROOM-1-N16R8`.

- Project monorepo: https://github.com/and3rson/lilka
- Docs (Ukrainian): https://lilka.readthedocs.io/uk/latest

We are expecting a lot of people to use it with `platformio`, thus having it in this repository would greatly simplify the set-up for new developers.

Thanks in advance, and please let me know if my configuration is missing anything!

**EDIT**: the reason for starting with  `v2` is that `v1` is an internal prototype that was demonstrated to research the feasibility of developing a stable model (`v2`). Thus the first public version is `v2`. So in order to avoid confusion, we're explicitly naming it `lilka_v2`.

Furthermore, `v3` is likely going to be developed in future being potentially backwards-incompatible, so including hardware version as part of board's name ensures long-term support for all versions.

### DIFF
--- a/boards/lilka_v2.json
+++ b/boards/lilka_v2.json
@@ -1,0 +1,60 @@
+{
+  "build": {
+    "arduino":{
+      "ldscript": "esp32s3_out.ld",
+      "memory_type": "qio_opi",
+      "partitions": "default_16MB.csv"
+    },
+    "core": "esp32",
+    "extra_flags": [
+      "-DARDUINO_LILKA",
+      "-DLILKA_VERSION=2",
+      "-DARDUINO_USB_MODE=1",
+      "-DARDUINO_USB_CDC_ON_BOOT=1",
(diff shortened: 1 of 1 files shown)


and3rson:
@valeros Thanks for the heads-up! This is a good question. I did my testing without `use_1200bps_touch`, and the entire flashing/rebooting process went without any issues, irregardless whether the board was already running another firmware or was just plugged in during "Looking for upload port...".

I think this might potentially be the case with my system specifically, or perhaps that's just the way the bootloader of ESP32-S3 works. If you suggest that adding `use_1200bps_touch` will make it potentially safer - then we should definitely add it!

and3rson:
@valeros Thank you so much!

PR #1295, opened 2024-01-29
Add board Nologo ESP32C3 Super Mini

imyelo:
refs:

- https://www.nologo.tech/product/esp32/esp32c3SuperMini/esp32C3SuperMini.html
- https://github.com/espressif/arduino-esp32/pull/9174

PR #1319, opened 2024-02-27
Fix flash speed for esp32-c3-devkitm-1

Cyberbeni:
I got this device for ESPresense: https://www.aliexpress.com/item/1005006399553064.html
When flashing the device with the stock ESPresence image for ESP32-C3, it gets stuck in a bootloop, adding this to platformio.ini and manually building fixed the issue for me (I saw these setting being recommended at multiple places, only f_flash is the problem according to [this](https://github.com/platformio/platform-espressif32/pull/623)):

```
board_build.f_flash = 40000000L
board_build.flash_mode = dio
board_build.flash_size = 4MB
```

### DIFF
--- a/boards/esp32-c3-devkitm-1.json
+++ b/boards/esp32-c3-devkitm-1.json
@@ -5,7 +5,7 @@
     },
     "core": "esp32",
     "f_cpu": "160000000L",
-    "f_flash": "80000000L",
+    "f_flash": "40000000L",
     "flash_mode": "qio",
     "mcu": "esp32c3",
     "variant": "esp32c3"


Jason2866:
The board you are referencing is a different one than the PR is made.

Jason2866:
@Cyberbeni Yes. First step is to provide support for the board in [ArduinoIDE repo](https://github.com/espressif/arduino-esp32).
After the board is added there, a PR can be made here to add this new board variant.

Jason2866:
Just saw, there is already #1295 for the super mini

PR #1329, opened 2024-03-14
Use GCC wrapped ar & ranlib to enable LTO

deece:
As per https://github.com/platformio/platform-ststm32/issues/386

Link Time Optimisation shifts compilation to the final link, allowing for more aggressive optimisations.

AR & RANLIB need to use their gcc equivalents to enable this, and will otherwise behave identically if LTO data is not in the objects.

(This should probably be done for all platforms)

### DIFF
--- a/builder/main.py
+++ b/builder/main.py
@@ -241,7 +241,7 @@ env.Replace(
     __get_board_flash_mode=_get_board_flash_mode,
     __get_board_memory_type=_get_board_memory_type,
 
-    AR="%s-elf-ar" % toolchain_arch,
+    AR="%s-elf-gcc-ar" % toolchain_arch,
     AS="%s-elf-as" % toolchain_arch,
     CC="%s-elf-gcc" % toolchain_arch,
     CXX="%s-elf-g++" % toolchain_arch,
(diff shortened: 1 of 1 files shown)


deece:
Note: This appears to a rehash of https://github.com/platformio/platform-espressif32/pull/702

Jason2866:
Using LTO since over a year with project Tasmota and it is working.
Knowing from maintainer of project EspEasy he is using too and it works well here too.

valeros:
The `espressif32` dev-platform is more popular than `ststm32`, so all steps regarding build system should be carefully considered. LTO in embedded builds is a controversial topic, so I'm a bit reluctant to enable it by default, mainly because of the aggressive optimizations as you mentioned. We try to differ as little as possible from the original tools used in frameworks, so if you still interested in enabling LTO for ESP32 targets, please file a feature request in the Espressif repositories in the first place.

Jason2866:
@valeros espressif Arduino is already using `elf-gcc-ar` https://github.com/espressif/arduino-esp32/pull/8013

valeros:
@Jason2866 thanks for the info. I didn't notice this change in 2.0.14, so I believe it's currently only available in the core dev branch for v3.x?

Jason2866:
@valeros Yes, it is added in dev branch v3.x and will be there in release v3.0.0

deece:
@valeros this change in itself doesn't enable LTO, it just removes things that get in the way of enabling it.

It should be completely transparent to non-LTO builds.

Jason2866:
Exactly, without explicit enabling LTO there will be no change in compiled code by merging the PR

valeros:
Thanks for the PR, merged!

PR #1334, opened 2024-03-19
Add missing Bluetooth connectivity for esp32-s3-devkitm-1

mb-orgatex:
esp32-s3-devkitm-1 was overseen when adding bluetooth in f6ec3926f9f660ee9abada8540ffe1e205da4bbf


valeros:
Thanks!

PR #1336, opened 2024-03-20
Handle "--encrypt" in upload_flags correctly for esptool

saschaklick:
The default way of applying upload_flags to the esptool arguments is not compatible with the way esptool expects "--encrypt" to be positioned.

Also upload compression is disabled when encrypted upload is requested.


valeros:
Hi @saschaklick, thanks for the PR. Currently, the firmware encryption workflow is not supported, but for your case I'd recommend using an [extra script](https://docs.platformio.org/en/latest/scripting/index.html) instead of `upload_flags`. This way you will be able to insert any additional flags at an arbitrary place in the upload command.

PR #1339, opened 2024-03-22
Add I2S pin defs to Lilka (v2) board

and3rson:
Apparently, there's no nice way to override I2S pins in ESP32 Arduino framework during runtime. So the best solution is to do this on a board level.


and3rson:
My bad, turns out I'm totally blind: there *are*, in fact, methods to do this.

Closing as not needed.

PR #1344, opened 2024-04-01
Add board ESP32-S3 PowerFeather

powerfeatherdev:
Adds support for ESP32-S3 PowerFeather (https://powerfeather.dev/). Board uses an 8MB flash, 2MB PSRAM ESP32-S3 module.

Board is supported since [Arduino ESP32 2.0.15 release](https://github.com/espressif/arduino-esp32/releases/tag/2.0.15), see merged PRs https://github.com/espressif/arduino-esp32/pull/9325 and https://github.com/espressif/arduino-esp32/pull/9398. Support PR's for the 3.x series has also been merged in https://github.com/espressif/arduino-esp32/pull/8889, https://github.com/espressif/arduino-esp32/pull/9052 and https://github.com/espressif/arduino-esp32/pull/9431 and should be in the next release.

Arduino variant is `esp32s3_powerfeather` with board name `ESP32S3_POWERFEATHER` (see https://github.com/espressif/arduino-esp32/blob/master/boards.txt#L35200-L35201)

ESP-IDF is also supported, but pin definitions are in the accompaniment library, [powerfeather-sdk](https://github.com/PowerFeather/powerfeather-sdk).

### DIFF
--- a/boards/esp32s3_powerfeather.json
+++ b/boards/esp32s3_powerfeather.json
@@ -1,0 +1,50 @@
+{
+  "build": {
+    "arduino":{
+      "partitions": "default_8MB.csv"
+    },
+    "core": "esp32",
+    "extra_flags": [
+      "-DARDUINO_ESP32S3_POWERFEATHER",
+      "-DARDUINO_USB_CDC_ON_BOOT=1",
+      "-DARDUINO_RUNNING_CORE=1",
+      "-DARDUINO_EVENT_RUNNING_CORE=1",
+      "-DBOARD_HAS_PSRAM"
+    ],
+    "f_cpu": "240000000L",
+    "f_flash": "80000000L",
+    "flash_mode": "qio",
+    "hwids": [
+      [
+        "0X303A",
+        "0x81BB"
+      ]
+    ],
+    "mcu": "esp32s3",
+    "variant": "esp32s3_powerfeather"
+  },
+  "connectivity": [
+    "bluetooth",
+    "wifi"
+  ],
+  "debug": {
+    "openocd_target": "esp32s3.cfg"
+  },
+  "frameworks": [
+    "arduino",
+    "espidf"
+  ],
+  "name": "ESP32-S3 PowerFeather",
+  "upload": {
+    "flash_size": "8MB",
+    "maximum_ram_size": 327680,
+    "maximum_size": 8388608,
+    "use_1200bps_touch": true,
+    "wait_for_upload_port": true,
+    "require_upload_port": true,
+    "speed": 460800
+  },
+  "url": "https://powerfeather.dev/",
+  "vendor": "PowerFeather"
+}
+
--- a/boards/esp32s3_powerfeather.json
+++ b/boards/esp32s3_powerfeather.json
@@ -8,8 +8,7 @@
       "-DARDUINO_ESP32S3_POWERFEATHER",
       "-DARDUINO_USB_CDC_ON_BOOT=1",
       "-DARDUINO_RUNNING_CORE=1",
-      "-DARDUINO_EVENT_RUNNING_CORE=1",
-      "-DBOARD_HAS_PSRAM"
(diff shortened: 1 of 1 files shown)


boards/esp32s3_powerfeather.json
```diff
@@ -0,0 +1,52 @@
+{
+  "build": {
+    "arduino":{
+      "ldscript": "esp32s3_out.ld",
```
Jason2866, line 4: ldscript entry is not used and existing since core 2.x

powerfeatherdev:
Hi @Jason2866 thank you for the input, as I am unfamiliar with adding new board support for PlatformIO. Is there a guide somewhere (also for adding support for a library: https://github.com/PowerFeather/powerfeather-sdk)?

Jason2866:
@powerfeatherdev Just add a library.json and/or library.properties in the repo and it can be used as lib with Platformio.

powerfeatherdev:
@Jason2866 For the `library.properties`, is it the [same one as the  Arduino library specification](https://arduino.github.io/arduino-cli/0.19/library-specification/)?

If so, [my library already has one](https://github.com/PowerFeather/powerfeather-sdk) when I added Arduino support - but in that case I had to submit it to the Arduino library registry repo. Does that mean all arduino libraries are discoverable by PlatformIO automatically?

Jason2866:
No, it is not accessing Arduino registry. But Platformio can use a github repo as lib source.
Uploading to Platformio registry is not needed to have access to the library.
It can be uploaded to load the lib from the Platformio registry. In this case a library.json has to be added to the package. https://docs.platformio.org/en/latest/core/userguide/pkg/cmd_pack.html

powerfeatherdev:
@Jason2866 I think I will submit the library to the PlatformIO registry - just makes it easier for the board's users.

Anyways, I resolved your initial comment and also filled in the PR description. PTAL. Thanks for answering my questions!

powerfeatherdev:
Oh, and one more question @Jason2866, is there a way to specify a board support on a library? I know for libraries you can do that through the `library.json` file, but for a board support json like this one, is there such a thing?

Jason2866:
@powerfeatherdev Not tried, nor does iknow if something like this exists.  If you add an entry after URL / Vendor? I added entrys in local boards.json for my use cases and it does not conflict.

powerfeatherdev:
Hi @Jason2866, that's fine. Just wondering if there was something like that.

powerfeatherdev:
@valeros PTAL as well.

powerfeatherdev:
@valeros, I assume this is waiting until https://registry.platformio.org/tools/platformio/framework-arduinoespressif32 gets updated to 2.0.15+?

valeros:
@powerfeatherdev That's right.

powerfeatherdev:
Hi @valeros, I see. Is there a timeline for updating it to 2.0.15+?

valeros:
> Hi @valeros, I see. Is there a timeline for updating it to 2.0.15+?

This or next week should be available in the dev branch.

valeros:
Hi @powerfeatherdev

I noticed that you enabled the `use_1200bps_touch` option in the board manifest, while in the `boards.txt` it's not set. Does this board need that touch procedure?

There is also a discrepancy in PSRAM settings, by default it's disabled in the original Arduino config, but in this PR you added the `BOARD_HAS_PSRAM` macro. 

Which settings are correct?

powerfeatherdev:
Hi @valeros, I copied the settings from Adafruit ESP32S3 Feather without double checking. Will fix this.

Jason2866:
Just my 2 cents. Before it was correct. Board has PSRAM. Arduino setup is wrong.

powerfeatherdev:
It seems the definition `BOARD_HAS_PSRAM` is assosciated with QSPI or OPI PSRAM being selected. Even though board has PSRAM, as I prefer to build without this by default - I'll leave it removed for now. @valeros @Jason2866

PR #1356, opened 2024-04-15
Add board definition for Wemos C3 Pico V1.0.0 - Lolin

lboue:
Add board WEMOS LOLIN C3 PICO https://www.wemos.cc/en/latest/c3/c3_pico.html


lboue:
Could you please tell me what is missing or validate this request?

PR #1361, opened 2024-04-19
Add board for Wireless Tag WT32-ETH02

darkxst:
WT32-ETH02 is a new variant based on a single core ESP32-SOLO-1 with 16MB flash.

### DIFF
--- a/boards/wt32-eth02.json
+++ b/boards/wt32-eth02.json
@@ -1,0 +1,36 @@
+{
+  "build": {
+    "core": "esp32",
+    "extra_flags": [
+      "-DARDUINO_WT32_ETH01",
+      "-DCONFIG_FREERTOS_UNICORE=1"
+    ],
+    "f_cpu": "240000000L",
+    "f_flash": "80000000L",
+    "flash_mode": "qio",
+    "mcu": "esp32",
+    "variant": "wt32-eth02"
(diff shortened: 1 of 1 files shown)


boards/wt32-eth02.json
```diff
@@ -0,0 +1,40 @@
...
+    "wifi",
+    "bluetooth",
+    "ethernet",
+    "can"
+  ],
+  "debug": {
+    "openocd_board": "esp32-solo-1.cfg"
+  },
+  "frameworks": [
+    "arduino",
```
Jason2866, line 27: Not compatible with Arduino. The precompiled static Arduino libs are for dual core esp32 only.
Every Esp32solo1 does not work with Arduino

boards/wt32-eth02.json
```diff
@@ -0,0 +1,40 @@
+{
+  "build": {
+    "arduino": {
```
Jason2866, line 3: remove entry Arduino and the linker script. Not used since Arduino core 2.0.0

Jason2866 (changes requested):

boards/wt32-eth02.json
darkxst: ok, have removed arduino.

PR #1362, opened 2024-04-20
change `build.esp-idf.preserve_source_file_extension` to `True`

Jason2866:
as default. It avoid issues like this https://github.com/espressif/esp-dsp/issues/74


valeros:
Thanks for the PR, merged.

PR #1377, opened 2024-05-14
Fix wt32-eth01.json

darkxst:
wt32-eth01 in esp-idf wont boot with qio flash mode and ethernet configured.

Some pins for QIO flash appear to be muxed with ethernet MAC. Given this a dedicated ethernet board, set flash mode back to 'dio'.

### DIFF
--- a/boards/wt32-eth01.json
+++ b/boards/wt32-eth01.json
@@ -9,7 +9,7 @@
     ],
     "f_cpu": "240000000L",
     "f_flash": "80000000L",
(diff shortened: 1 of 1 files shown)


Jason2866:
The original Module does work with flash mode QIO.

darkxst:
> The original Module does work with flash mode QIO. There are many bad clones sold.

I get these boards direct from Wireless-Tag so they are not clones. QIO works fine until I enable the ethernet MAC, then the board wont boot.

PR #1391, opened 2024-06-06
Create esp32-c3-supermini

mvoss96:
Add Board definition for ESP32-C3 supermini

### DIFF
--- a/boards/esp32-c3-supermini
+++ b/boards/esp32-c3-supermini
@@ -1,0 +1,35 @@
+{
+  "build": {
+    "arduino":{
+      "ldscript": "esp32c3_out.ld"
+    },
+    "core": "esp32",
+    "f_cpu": "160000000L",
+    "f_flash": "80000000L",
+    "flash_mode": "qio",
+    "extra_flags": "-DARDUINO_ESP32C3_DEV",
+    "mcu": "esp32c3",
+    "variant": "esp32c3"
+  },
+  "connectivity": [
+    "wifi",
+    "bluetooth"
+  ],
+  "debug": {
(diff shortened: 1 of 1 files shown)


Jason2866:
Not needed since the board `adafruit_qtpy_esp32c3.json` fits 100%
Beside of this your definition is missing the CDC settings.

mak0t0san:
> Not needed since the board `adafruit_qtpy_esp32c3.json` fits 100% Beside of this your definition is missing the CDC settings.

Good to know as I also have the ESP32-C3 Supermini and ended up creating my own board definition. I was about to make a PR to add it but then saw there was already one and came across the comment of using `adafruit_qtpy_esp32c3` . It would be super helpful if this information was more accessible. How were you able to determine that the supermini and adafruit_qtpy_esp32c3 were the same? This would be a helpful process to know for me in the future when I come across a board that I cannot find a match for (by name).

yanxke:
> > Not needed since the board `adafruit_qtpy_esp32c3.json` fits 100% Beside of this your definition is missing the CDC settings.
> 
> Good to know as I also have the ESP32-C3 Supermini and ended up creating my own board definition. I was about to make a PR to add it but then saw there was already one and came across the comment of using `adafruit_qtpy_esp32c3` . It would be super helpful if this information was more accessible. How were you able to determine that the supermini and adafruit_qtpy_esp32c3 were the same? This would be a helpful process to know for me in the future when I come across a board that I cannot find a match for (by name).

Agree with this.  Can we at least modify the Adafruit board description to include these keywords so that it shows up in the search, since the supermini is a much more popular and accessible board.

PR #1394, opened 2024-06-07
Add board viralink full

armp30:
Add ViraLink Full Connectivity Board that has on board lan8720 sim800x max485 and Wiegand interface


valeros:
Hi @armp30, there is no such variant `viralink_full` in the official Arduino repo for ESP32. You need to add a proper variant to https://github.com/espressif/arduino-esp32 in the first place.

PR #1399, opened 2024-06-13
Added Board definition for M5Stack Core Basic 16MB variant

TheMasterofBlubb:
https://docs.m5stack.com/en/core/basic

According to the documentation (see Version Change) around 2020 the M5Stack Core Basic got upgraded to 16MB Flash. This also covers the newer versions with the SKUs K001-V27 and K001-V26.

### DIFF
--- a/boards/m5stack-core-esp32-16M.json
+++ b/boards/m5stack-core-esp32-16M.json
@@ -1,0 +1,34 @@
+{
+  "build": {
+    "arduino":{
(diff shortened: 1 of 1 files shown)


TheMasterofBlubb:
Note: i left the old board definition as is to not mess with backwards compatibility

valeros:
Thanks for the PR, added.

PR #1414, opened 2024-07-02
Enable PSRAM for adafruit_feather_esp32s3_reversetft

pcorliss:
This board [ships with 2Mb of PSRAM](https://learn.adafruit.com/esp32-s3-reverse-tft-feather/overview). Enabling by default.


valeros:
Thanks for the PR, merged.

mariusGundersen:
This should have been done to the non-reverse tft board as well. Is it enough to do what was done in this pull request, or is the `"memory_type": "qio_qspi"` line also needed?

PR #1426, opened 2024-07-15
add littlefs partition subtype

dariush:
Added "littlefs" to list of recognised subtypes in  _parse_partitions so that partition table configurations using littleFS build successfully.
See https://community.platformio.org/t/build-filesystem-image-fails-for-littlefs-partition/41709


valeros:
Thanks, merged!

PR #1429, opened 2024-07-25
Add Freenove ESP32-WROVER(-E) board

oliv3r:
The Freenove [0] ESP32-WROVER board comes with 4MiB flash and 8MiB of PSRAM in the ESP32-WROVER-E. It has been around for a while and clones also exist. Usually with the silk screen modified to `ESP32-WROVER-DEV`. The board has a camera connector, but sadly no onboard flash, like other camera boards.

The board also has 4 LED's, A green power LED, amber TX/RX leds connected to the UART0 pins and a blue status led connected on GPIO02.

The AI Tinkerer cam32 board was used as reverence for writing this definition, as that is what I am currently using with this module.

[0]: https://store.freenove.com/products/fnk0060


valeros:
Thanks, added.

PR #1434, opened 2024-07-30
Fix fail for IDF < v5.3

Jason2866:
latest changes breaks function for IDF 5 minor versions < 5.3


valeros:
Thanks for the PR, merged with some changes to version check approach.

Jason2866:
Thx, your changes are from a real Python coder :-)
I am a Python noob...

PR #1441, opened 2024-08-13
Add board def for Adafruit Feather ESP32-C6

tyeth:
This adds the Adafruit Feather ESP32-C6 (4mb variant no psram).

About to test it and then will make non-draft. Also need to confirm the board define is correct.

Should match up to https://github.com/espressif/arduino-esp32/tree/master/variants/adafruit_feather_esp32c6

### DIFF
--- a/boards/adafruit_feather_esp32c6.json
+++ b/boards/adafruit_feather_esp32c6.json
@@ -1,0 +1,47 @@
+{
+  "build": {
+    "arduino":{
+      "ldscript": "esp32c6_out.ld"
+    },
+    "core": "esp32",
+    "extra_flags": [
+      "-DARDUINO_ADAFRUIT_FEATHER_ESP32C6",
+      "-DARDUINO_USB_MODE=1",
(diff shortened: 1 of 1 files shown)


valeros:
Hi @tyeth, thanks for the PR. This board was added in the Arduino core v3.0.3 which is currently unsupported, see #1225.

tyeth:
Thanks @valeros 
I did see that this would be the case last night when looking at another board PR that needs 3.0.3 and was left open.

Guess this will sit here for a while, but it's possibly useful for our testing (and others).

tyeth:
For anyone trying to work with the latest processors in Arduino-ESP32 board support package v3, you can use the community maintained fork:
https://github.com/pioarduino/platform-espressif32/releases/

So in your esp32 board related platformio.ini `[env]` section, use the appropriate zip from a versioned or latest release (or #dev branch):
```
[env:featherc6]
platform = https://github.com/pioarduino/platform-espressif32/releases/latest/download/platform-espressif32.zip
board = adafruit_feather_esp32c6
```

PR #1469, opened 2024-09-28
Add board definition for Freenove ESP32-S3-WROOM

argetlam-coder:
Adds support for Freenove ESP32-S3-WROOM (https://github.com/Freenove/Freenove_ESP32_S3_WROOM_Board)


valeros:
Thanks for the PR, added!

PR #1480, opened 2024-10-24
Update heltec_wireless_stick_lite.json

Andreas-Pedersen:
Supports 80MHz
I get an error when using this board and a fresh install of platformio in vscode, change this speed to 80Mhz solves the problems.


valeros:
Thanks, merged.

PR #1481, opened 2024-10-24
Update heltec_wireless_stick_lite.json

Andreas-Pedersen:
Changed the fields so that they match an esp32-s3fn8 that is on the heltec-wireless-stick-lite board.


valeros:
Hi @Andreas-Pedersen, you're probably referring to a newer version of this board called `Wireless Stick Lite (V3)`. The board manifest you've changed is meant for an older version based on a generic ESP32. I'd recommend creating a new board manifest `heltec_wireless_stick_lite_v3.json` with proper settings for the ESP32-S3 target.

PR #1489, opened 2024-11-15
Add RYMCU ESP32-S3-DevKitC-1 board

ronger-x:


ronger-x:
> Hi @ronger-x, thanks for the PR. There is probably a mistake in the variant field `"variant": "esp32s3n8r2"`. There is no `esp32s3n8r2` option available, you can use `esp32s3` instead.

Thank you, the error has been corrected.

valeros:
Thanks, merged.

PR #1498, opened 2024-12-02
Fixed missing psram for adafruit feather with tft

mariusGundersen:
This is the same config as for the reverse tft


valeros:
Thanks, merged!

PR #1512, opened 2024-12-11
Add USB_CDC flag to Romeo ESP32-S3

PabloPL:
Adding "-DARDUINO_USB_CDC_ON_BOOT=1" to dfrobot_romeo_esp32s3.json, otherwise, the serial monitor from USB doesn't work.

Similar to https://github.com/platformio/platform-espressif32/pull/888


PabloPL:
By the way, maybe we could also mention that with esp-bridge this device is debuggable using Platformio, without any other HW?  On Wiki there is following information
```DFRobot Romeo ESP32-S3 does not have on-board debug probe and IS NOT READY for debugging. You will need to use/buy one of external probe listed below.```

PR #1536, opened 2025-01-31
Add Shelly3Pro

oliv3r:
Add the Shelly3Pro, as found in the Shelly Pro 3. This is a custom designed ESP32 board with 8MiB flash, and SMSC 8720A Ethernet PHY. All other I/O is on separate boards via pin-headers and thus not relevant.

The board has no flashing circuity and requires 'manual' toggling of reset and GPIO0.

### DIFF
--- a/boards/shelly3pro.json
+++ b/boards/shelly3pro.json
@@ -1,0 +1,37 @@
+{
+  "build": {
+    "arduino": {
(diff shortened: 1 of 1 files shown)


Jason2866:
There is no `shelly3pro_esp32` board variant in any Arduino core.

PR #1588, opened 2025-05-17
use provided upload port

### LINKED ISSUE #1582: "Upload FileSystem Image" use unexpected port
I have two ESP32 connected, two ports COM26, COM49 are available to upload. Next I change "auto" selecting port to COM49 for instance, but when I hit "upload file system image" I see that auto scan had happen, and unexpected port was used to upload. Following screen show that:
Image

 I expect then if I do specify port explicitly it shopuld be used to "upload filesystem image", similar to just "upload" command.
Thanks in advance.

Jason2866:
and if none found use auto detection for the port. Partially fixes #1582 
When the port is provided via Platformio VSC extension it is still auto detected. Dont know how to get choosen settings from the Platformio VSC extension.

### DIFF
--- a/builder/main.py
+++ b/builder/main.py
@@ -32,7 +32,8 @@ def BeforeUpload(target, source, env):
     if "BOARD" in env:
         upload_options = env.BoardConfig().get("upload", {})
 
-    env.AutodetectUploadPort()
+    if not env.subst("$UPLOAD_PORT"):
+        env.AutodetectUploadPort()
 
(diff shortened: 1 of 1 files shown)


valeros:
Hi @Jason2866, thanks for the PR. Unfortunately, your PR won't fix the issue as it spans across different parts of PlatformIO ecosystem (changes also required for the IDE extension and Core in the first place).

PR #1597, opened 2025-06-01
support esp32c5

mischief:
here's my attempt at supporting the esp32c5 mcu and esp32-c5-devkitc-1, which i ordered from the [espressif official online store] and arrived recently.

i've only attempted to work on espidf framework, since i am pretty sure there would need to be upstream work done to support esp32c5 in the arduino core.

note that i am a total n00b at python, so i apologize in advance.

[espressif official online store]: https://www.aliexpress.us/item/3256808604473710.html

### DIFF
--- a/builder/frameworks/_embed_files.py
+++ b/builder/frameworks/_embed_files.py
@@ -110,14 +110,14 @@ env.Append(
                 " ".join(
                     [
                         "riscv32-esp-elf-objcopy"
-                        if mcu in ("esp32c3", "esp32c6")
+                        if mcu in ("esp32c3", "esp32c5", "esp32c6")
                         else "xtensa-%s-elf-objcopy" % mcu,
                         "--input-target",
                         "binary",
                         "--output-target",
-                        "elf32-littleriscv" if mcu in ("esp32c3","esp32c6") else "elf32-xtensa-le",
+                        "elf32-littleriscv" if mcu in ("esp32c3", "esp32c5", "esp32c6") else "elf32-xtensa-le",
                         "--binary-architecture",
-                        "riscv" if mcu in ("esp32c3","esp32c6") else "xtensa",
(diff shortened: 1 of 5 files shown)


builder/frameworks/_embed_files.py
```diff
@@ -110,14 +110,14 @@ def transform_to_asm(target, source, env):
                 " ".join(
                     [
                         "riscv32-esp-elf-objcopy"
-                        if mcu in ("esp32c3", "esp32c6")
+                        if mcu in ("esp32c3", "esp32c5", "esp32c6")
```
mischief, line 113: this (and the following) repeated lists of mcu feels like it could be refactored into each board json, or at least centralized into a function that said whether the mcu is riscv32 or not.

mischief:
esp32c5 does not work yet with older tool-esptoolpy 1.40501.0, but does kind of work with newer 1.40801.0.

Jason2866:
you can use the fork pioarduino https://github.com/pioarduino/platform-espressif32/tree/develop_33

mischief:
> you can use the fork pioarduino https://github.com/pioarduino/platform-espressif32/tree/develop_33

@Jason2866 thanks! this appears to work for the ESP32-C5-DevKitC-1. is there a plan to merge this back to the platformio repos?

your repo has a bug, or was maybe was brought up against a different SoC. my ESP32-C5-DevKitC-1 v1.2 contains the ESP32-C5-WROOM-1-N8R4 (doc here https://www.espressif.com/sites/default/files/documentation/esp32-c5-wroom-1_wroom-1u_datasheet_en.pdf), which has 8MB flash and 4MB PSRAM. your config says 4MB flash and no PSRAM.

Jason2866:
@mischief 
> Jason2866 thanks! this appears to work for the ESP32-C5-DevKitC-1. is there a plan to merge this back to the platformio repos?

No, see https://github.com/platformio/platform-espressif32/issues/1225

PR #1604, opened 2025-06-23
Add new board M5Stack StampC3

joshdoesthis:
PlatformIO board configuration for the M5Stack StampC3.


joshdoesthis:
I believe the variant for this board can be set to `m5stack_stamp_c3` once [platformio/framework-arduinoespressif32](https://registry.platformio.org/tools/platformio/framework-arduinoespressif32) has the required details from [espressif/arduino-esp32](https://github.com/espressif/arduino-esp32/tree/master).

Issue #1569, opened 2025-04-16
Add board support for Waveshare ESP32-S3-ETH


danieljharris:
Waveshare page: [https://www.waveshare.com/wiki/ESP32-S3-ETH](https://www.waveshare.com/wiki/ESP32-S3-ETH)
PiHut listing: [https://thepihut.com/products/esp32-s3-ethernet-development-board-rj45](https://thepihut.com/products/esp32-s3-ethernet-development-board-rj45)

![Image](https://github.com/user-attachments/assets/58fa45a0-3ef2-48cd-899d-3209450cd9a3)

Fabimawn:
Is there a config we can use for it in the meanwhile? I need to work with it within a deadline.

valeros:
@Fabimawn There are several boards with similar specs, for example `4d_systems_esp32s3_gen4_r8n16`.

Fabimawn:
Fantastic, thats what I was looking for! I just didn't see a config that fit this ESP specifically so far since most of my boards were the WROOM variants

PR #1652, opened 2025-09-18
Add new board ESP32-S3-ETH

danieljharris:
Add support for: https://www.waveshare.com/wiki/ESP32-S3-ETH

Fixes Issue: https://github.com/platformio/platform-espressif32/issues/1569

Recommended ESP32-S3-ETH config (from their site):
image

### DIFF
--- a/boards/waveshare_esp32s3_eth.json
+++ b/boards/waveshare_esp32s3_eth.json
@@ -1,0 +1,51 @@
+{
+  "build": {
+    "arduino": {
+      "ldscript": "esp32s3_out.ld",
+      "memory_type": "qio_opi",
+      "partitions": "huge_app.csv"
+    },
+    "core": "esp32",
+    "extra_flags": [
+      "-DARDUINO_ESP32_S3R8N16",
+      "-DBOARD_HAS_PSRAM",
+      "-DARDUINO_USB_MODE=1",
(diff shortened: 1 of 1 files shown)


valeros:
Hi @danieljharris, thanks for the PR. I see `ESP32S3 Dev Module` is used as the board on your screenshot , but in the manifest you set `esp32s3eth` in the `variant` field.

amayhall-gloop:
Any update on this making it into the board explorer? No rush I was just curious as I've got a couple projects using these boards already and wanted to make the jump into PlatformIO.

danieljharris:
> ESP32S3 Dev Module

From other examples such as [esp32s3camlcd](https://github.com/platformio/platform-espressif32/blob/develop/boards/esp32s3camlcd.json) it looks like this approach to variant names is common. Please correct me if I am wrong and there is a better name for this. `esp32s3eth` better describes the variant than just dev module as there are specific configs for this variant of board recommended by the manufacturer.

alexmiller422:
When using this board file with the aruino framework, I get `fatal error: pins_arduino.h: No such file or directory`

I believe this file needs to exist under the directory `.platformio/packages/framework-arduinoespressif32/variants/<variant name>`

Setting the vairnat to `esp32s3` works.

PR #1701, opened 2026-01-09
Uploadfs to specific partition

rivimey:
In preparation, update parse_partitons() to print the partition table as parsed, which may differ from that in the file under some circumstances, and also to print a warning if there are too few tokens (which causes a "partition" to be ignored).

This in aid of making it possible for platformio.ini to specify exactly which partition the uploadfs action should write to, and making it clear in the logs which has been selected.

The board_build flags (esp the one I add here: fs_label) should be documented somwehere but I can't see where to put it?

Aside: The issue of writing to the wrong fs partition, with no obvious markers as to why, has caused me no end of pain, hence bothering to create & submit this patch.

Background: I have a partition scheme with 2 app partitions (factory, ota_0), two filesystem partitions (factory_fs, ota_0_fs), and a user filesystem (user_fs), the logic being that the first 4 of those were writable on FCS or by an OTA update and the last was for user data and so not updatable. You can't share factory_fs and ota_0_fs because then the code may break if you do "factory reset". So if you need filesystem with your app, you need different ones for both factory and ota. For me, right now, I don't need user_fs but it's there in case I do. As you will know, the existing logic in fetch_fs_size caused "user_fs" to be selected for any uploadfs action, which was definitely not what I wanted!

### DIFF
--- a/builder/main.py
+++ b/builder/main.py
@@ -131,13 +131,17 @@ def _parse_partitions(env):
     # The first offset is 0x9000 because partition table is flashed to 0x8000 and
     # occupies an entire flash sector, which size is 0x1000
     next_offset = 0x9000
+    linenum = 0
     with open(partitions_csv) as fp:
         for line in fp.readlines():
+            linenum += 1
             line = line.strip()
             if not line or line.startswith("#"):
                 continue
             tokens = [t.strip() for t in line.split(",")]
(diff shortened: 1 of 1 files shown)


rivimey:
Bump! I find this invaluable...

PR #1724, opened 2026-03-19
Update flash size and maximum size for Edgebox-ESP-100 to 16MB

simon-spobu:
See https://www.seeedstudio.com/EdgeBox-ESP-100-p-5490.html


valeros:
Many thanks, merged.

